PR #1, opened 2019-08-17
Add nm-vpn, battery, wifi and keyboard layout blocklet

leosunmo:
image

This adds a few new blocklets to enhance the default set of blocklets.

Not sure about the colour on the Battery2 blocklet yet, but it's at least a handy visual guide for myself to remember to charge my laptop.
The Battery2 blocklet also supports multiple batteries which is something the default Battery blocklet doesn't.

### DIFF
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -1,6 +1,5 @@
 # i3xrocks config file
-#
-# The man page for i3blocks is hosted at http://vivien.github.io/i3blocks
+# # The man page for i3blocks is hosted at http://vivien.github.io/i3blocks
 #
 # List of valid properties:
 #
@@ -49,9 +48,10 @@ min_width=100%
 
 # Battery indicator
 #
-# The battery instance defaults to 0.
-[battery]
-label= 
+# Displays an icon representing current charge as well as
+# a coloured percentage of battery charge remaining.
+# It also displays the estimated time left on current charge.
+[battery2]
 interval=30
 
 # Date Time
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -28,6 +28,15 @@ separator_block_width=40
 markup=pango
 color=xresource:i3xrocks.label.color
 
+# Wifi
+#
+# Displays the wifi network's name
+# If the auto-discovered wifi device is incorrect,
+# override the "instance" value with the desired device
+[wifi2]
+#instance=wlp3s0 Replace with your wifi device if you need
+interval=5
+
 # CPU usage
 #
 # The script may be called with -w and -c switches to specify thresholds,
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -27,6 +27,13 @@ separator_block_width=40
 markup=pango
 color=xresource:i3xrocks.label.color
 
+# Network manager VPN
+#
+# Displays currently connected VPN using the nm-cli tool.
+#[nm-vpn]
+#label=VPN:
+#interval=5
+
 # Wifi
 #
 # Displays the wifi network's name
@@ -35,6 +42,12 @@ color=xresource:i3xrocks.label.color
 [wifi2]
 #instance=wlp3s0 Replace with your wifi device if you need
 interval=5
+
+# Keyboard layout
+#
+# Displays current keyboard layout.
+#[keyboard_layout]
+#interval=persist
 
 # CPU usage
 #
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -48,11 +48,14 @@ min_width=100%
 
 # Battery indicator
 #
-# Displays an icon representing current charge as well as
-# a coloured percentage of battery charge remaining.
-# It also displays the estimated time left on current charge.
+# Displays total percentage charge left across all batteries.
+# Can optionally display time left with display_time_left (value doesn't matter).
+# Can also change the color of the percentage text depending on
+# charge remaining with percent_color (value doesn't matter).
 [battery2]
 interval=30
+#display_time_left=true
+percent_color=true
 
 # Date Time
 #
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -49,13 +49,18 @@ min_width=100%
 # Battery indicator
 #
 # Displays total percentage charge left across all batteries.
-# Can optionally display time left with display_time_left (value doesn't matter).
-# Can also change the color of the percentage text depending on
-# charge remaining with percent_color (value doesn't matter).
+#
+# To display colored text (or icon, if "icon_only" is enabled)
+# depending on charge remaining, uncomment "percent_color".
+#
+# To display est. time remaining, uncomment "display_time_left".
+#
+# To display the battery icon only, uncomment "icon_only".
 [battery2]
 interval=30
 #display_time_left=true
 percent_color=true
+#icon_only=true
 
 # Date Time
 #
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -25,13 +25,13 @@
 command=/usr/share/i3xrocks/$BLOCK_NAME
 separator_block_width=40
 markup=pango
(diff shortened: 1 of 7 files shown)


leosunmo:
If someone could test the behaviour of the keyboard_layout blocklet to let me know what it displays if you only have 1 layout, that'd be great!

leosunmo:
I'm not very well versed in Python, so if there's some improvements to be made, feel free to change it.

leosunmo:
Ugh, the battery script uses ACPI which is pretty much deprecated. We need to start using `upower` which is a lot trickier to parse out of the box. Ideally we'd consume it straight from the dbus. Might look in to building a static Go binary that can be run as a daemon.

kgilmer:
I tested the keyboard_layout blocket as requested.  When adding to the bar with the following config, I see no change from before:
```
[keyboard_layout]
label=K
interval=5
```

When I run the script on the command line I get this:
```
$ ./keyboard_layout
@a(ss) []
```

leosunmo:
Hm, might have to fix that. Worst case it can be opt-in I guess.

kgilmer:
I notice from the screenshot that the font is mixed between Source Code Pro and (I'm guessing) FontAwesome.  Each blocklet has to specify the font otherwise the wrong one is used by i3bar.  Here's an example:
```
printf "<span font_desc='%s' color='%s'>%2d%%</span>\n", `xrescat i3xrocks.value.font`, `xrescat i3xrocks.value.color`, $cpu_usage;
```
Not a great example, Perl and all.

In essence the blocklet needs to ouput:
```
<span font_desc='Source Code Pro Medium 13' color='#D8DEE9'> some text</span>
```

With the font and color coming from Xresources.

kgilmer:
I think opt-in for keyboard layout is fine.  I cannot see a usecase for using that if you only have one keyboard layout.  What do you think?

leosunmo:
Added commented out examples for keyboard layout and network manager VPN.

leosunmo:
Regarding the Font output, I thought it would default to what was in the i3xrocks.conf file?

leosunmo:
Okay, not the tidiest commit, but I spent too much time hacking on this and forgetting to commit.

Blocklets now adhere to some more sane config hierarchies.

`~/.config/i3xrocks/i3xrocks.conf`:
```
# Global properties
#
# The top properties below are applied to every block, but can be overridden.
# Each block command defaults to the script name to avoid boilerplate.
command=/usr/share/i3xrocks/$BLOCK_NAME
separator_block_width=40
markup=pango
color=xresource:i3xrocks.value.color
label_color=xresource:i3xrocks.label.color
<...>
```
These are the global defaults that all Blocklets should use (fallback to `xrescat` in the scripts).
This means you can manually override them in the i3xrocks config.

leosunmo:
Manual override for testing:
image

Falling back to my Xresources if no color is set in the config:
image

leosunmo:
And with some new Nord values for `i3xrocks` Xresources:
image

kgilmer:
This is all a fantastic effort.  Making the color values consistent and cleanup up a lot of cruft is much appreciated.  I think we need to specify the font as well as the color though.  It's hard for me to be sure without seeing the whole bar in your screenshots but I think the value font is not Source Code Pro.  I guess maybe you're not expecting that if you've changed your font to something else, so let me explain from my machine.  After taking your latest changes onto my machine, I see this:

Screenshot from 2019-08-17 08-39-46

You'll notice that the value font is pretty large.  I don't know what that font is but it's not Source Code Pro 13 Medium, which is what it's supposed to be as set in the i3bar's stanza in the i3 config file.

If I make these changes to `wifi2` and `cpu_usage` :
```diff
diff --git a/i3xrocks.conf b/i3xrocks.conf
index 43c00fe..d2b6701 100644
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -26,8 +26,10 @@ command=/usr/share/i3xrocks/$BLOCK_NAME
 separator_block_width=40
 markup=pango
 color=xresource:i3xrocks.value.color
+font=xresource:i3xrocks.value.font
 label_color=xresource:i3xrocks.label.color
 
+
 # Network manager VPN
 #
 # Displays currently connected VPN using the nm-cli tool.
diff --git a/scripts/cpu_usage b/scripts/cpu_usage
index 765d210..4535b3f 100755
--- a/scripts/cpu_usage
+++ b/scripts/cpu_usage
@@ -44,9 +44,10 @@ $cpu_usage eq -1 and die 'Can\'t find CPU information';
 # Get color values
 my $label_color = $ENV{'label_color'} || `xrescat i3xrocks.label.color`;
 my $value_color = $ENV{'color'} || `xrescat i3xrocks.value.color`;
+my $value_font = $ENV{'font'} || `xrescat i3xrocks.value.font`;
 
 # Print short_text, full_text
-printf "<span color='%s'></span><span color='%s'>%2d%%</span>\n", $label_color, $value_color, $cpu_usage;
-printf "<span color='%s'>%2d%%</span>\n", $value_color, $cpu_usage;
+printf "<span color='%s'></span><span font_desc='%s' color='%s'>%2d%%</span>\n", $label_color, $value_font, $value_color, $cpu_usage;
+printf "<span font_desc='%s' color='%s'>%2d%%</span>\n", $value_font, $value_color, $cpu_usage;
 
 exit 0;
diff --git a/scripts/wifi2 b/scripts/wifi2
index 9cf84a6..d434361 100755
--- a/scripts/wifi2
+++ b/scripts/wifi2
@@ -5,6 +5,7 @@ DEFAULT_DEVICE=$(cat /proc/net/wireless | perl -ne '/(\w+):/ && print $1')
 INTERFACE="${BLOCK_INSTANCE:-${DEFAULT_DEVICE}}"
 
 VALUE_COLOR=${color:=$(xrescat i3xrocks.value.color)}
+VALUE_FONT=${font:=$(xrescat i3xrocks.value.font)}
 LABEL_COLOR=${label_color:=$(xrescat i3xrocks.label.color)}
 
 
@@ -16,5 +17,5 @@ LABEL_COLOR=${label_color:=$(xrescat i3xrocks.label.color)}
 ESSID=$(/sbin/iwconfig $INTERFACE | perl -n -e'/ESSID:"(.*?)"/ && print $1')
 
 
-echo "<span color=\"${LABEL_COLOR}\"> </span><span color=\"${VALUE_COLOR}\">$ESSID</span>" # full text
-echo "<span color=\"${VALUE_COLOR}\">$ESSID</span>" # short text
+echo "<span color=\"${LABEL_COLOR}\"> </span><span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\">$ESSID</span>" # full text
+echo "<span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\">$ESSID</span>" # short text
```
Here is the result:
Screenshot from 2019-08-17 08-59-52

You can see that the font is now SCP for the modified blocks.

kgilmer:
The most concise way I can think of describing the bug is, the text font in i3bar (workspace index numbers on the left) should be the same font as i3xrocks blocklets of text on the right.

leosunmo:
Yeah I'll tackle the font next. Mostly focused on colour first as it was a bit tricky with the two values and separate languages used. 
Now that I've figured out a pattern to follow, adding font should be easy.

leosunmo:
Fonts should work now, please do give it a try if you can, I've made so many modifications to my system that I don't want to make incorrect assumptions! :P

kgilmer:
Fonts are now SCP, nice work.  I noticed that battery2 crashed on my thinkpad X1:
Screenshot from 2019-08-17 17-28-57

leosunmo:
Ugh, I know what I've done. Let me fix it.

kgilmer:
I'll merge this and push a release to regolith-unstable now for user testing.

kgilmer:
regolith-i3xrocks-config 1.10-1ubuntu1ppa1 is currently building in regolith-unstable.

PR #6, opened 2019-09-17
Update Wifi2 blocklet to use nmcli

leosunmo:
fixes regolith-linux/regolith-desktop#122

This is using a more modern commandline tool for network device discovery and management.
Works nicely with UTF-8 and requires less messing around in `/proc/` and `/sys/`

It does have this weird delay when it runs for the first time in a while, but not sure it matters much.


kgilmer:
Thanks @leosunmo !  In addition to the fix, very happy for the overall reduction in code.

PR #7, opened 2019-10-04
add wireguard support and fancy icon

gigglearrows:
This one did not work for me using nordvpn. since it is called :tun:tun0 instead of :vpn:tun0. I also added support for wireguard and added a fancy icon :) Note that we must update the .Xresources file to add a green color to it in the i3xblock.vpn.color variable. (or just default grey color if we rather want that. I use a green color so that it is easier to see that im in "secure" mode)

### DIFF
--- a/scripts/nm-vpn
+++ b/scripts/nm-vpn
@@ -4,8 +4,8 @@ CONNECTION=$(nmcli -t connection show --active | awk -F ':' '
 /wg0|tun0/{vpn="ON"}; {if(/wireguard|vpn/) {vpn_name=$1} else if (/tun0/){tun_name=$1}}
 END{if ( (vpn) && (vpn_name)) {printf("%s", vpn_name)}  else if( (vpn) && (tun_name)) {printf("%s", tun_name)}}')
 
-LABEL_COLOR=${vpn_label_color:-$(xrescat i3xrocks.vpn.label_color)}
-VALUE_COLOR="${vpn_color:-$(xrescat i3xrocks.vpn.color)}"
+LABEL_COLOR=${vpn_label_color:-$(xrescat i3xrocks.label.color)}
+VALUE_COLOR="${vpn_color:-$(xrescat i3xrocks.value.color)}"
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font)}
 
 if [[ -z "$CONNECTION" ]]; then
--- a/scripts/nm-vpn
+++ b/scripts/nm-vpn
@@ -4,7 +4,7 @@ CONNECTION=$(nmcli -t connection show --active | awk -F ':' '
 /wg0|tun0/{vpn="ON"} /wireguard|tun|vpn/{name=$1}
 END{if(vpn) printf("%s", name)}')
 
-LABEL_COLOR=${vpn_color:-$(xrescat i3xrocks.vpn.color)}
+LABEL_COLOR=${vpn_label_color:-$(xrescat i3xrocks.vpn.label_color)}
 VALUE_COLOR="${vpn_color:-$(xrescat i3xrocks.vpn.color)}"
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font)}
 
--- a/scripts/nm-vpn
+++ b/scripts/nm-vpn
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 CONNECTION=$(nmcli -t connection show --active | awk -F ':' '
-/wg0|tun0/{vpn="ON"} /wireguard|tun0|vpn/{name=$1}
-END{if(vpn) printf("%s", name)}')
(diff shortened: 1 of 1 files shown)


kgilmer:
LGTM.  I can handle adding the vpn color to regolith-styles.

kgilmer:
@leosunmo any concerns?

leosunmo:
This is displaying `tun0` for me instead of the actual VPN name. I'm connecting over OpenVPN and `nmcli` is displaying both `tun0` as well as a `vpn` with the name of my connection.

```
$  nmcli -t connection show --active
leo.palmer:a66c233a-fb83-4ca3-a510-fe3ab571933a:vpn:wlp3s0
It hertz when IP:8a34dae9-9dc5-47bc-8ef6-61f4501898fb:802-11-wireless:wlp3s0
tun0:c80de262-9a81-403d-bc55-39af1bdb68f1:tun:tun0
```

So we need to figure out a better way to select the correct label to display.

Does NordVPN not add a named connection in the list from `nmcli -t connection show --active` for you @gigglearrows ?

scripts/nm-vpn
```diff
@@ -1,16 +1,16 @@
 #!/bin/bash
 
 CONNECTION=$(nmcli -t connection show --active | awk -F ':' '
-/tun0/{vpn="ON"} /vpn/{name=$1}
+/wg0|tun0/{vpn="ON"} /wireguard|tun0|vpn/{name=$1}
 END{if(vpn) printf("%s", name)}')
```
leosunmo: Try this instead:
```
CONNECTION=$(nmcli -t connection show --active | awk -F ':' '
/wg0|tun0/{vpn="ON"}; {if(/wireguard|vpn/) {vpn_name=$1} else if (/tun0/){tun_name=$1}}
END{if ( (vpn) && (vpn_name)) {printf("%s", vpn_name)}  else if( (vpn) && (tun_name)) {printf("%s", tun_name)}}')
```
This will prioritise `wireguard` and `vpn` names and fall back to `tun0` names.
Quite verbose, but performant enough and it'll make it easier to read for others.

Let me know if it works with your VPN client.

scripts/nm-vpn
```diff
@@ -1,16 +1,16 @@
...
+LABEL_COLOR=${vpn_label_color:-$(xrescat i3xrocks.vpn.label_color)}
+VALUE_COLOR="${vpn_color:-$(xrescat i3xrocks.vpn.color)}"
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font)}
 
 if [[ -z "$CONNECTION" ]]; then
     exit 0
 else
-   echo "<span font_desc=\"${VALUE_FONT}\" color=\"${LABEL_COLOR}\">vpn: </span><span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\">$CONNECTION</span>" # full text
+   echo "<span color=\"${LABEL_COLOR}\" font=\"FontAwesome\"> </span> <span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\">$CONNECTION</span>" # full text
    echo "<span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\">ON</span>" # short text
```
leosunmo: We should change this to just show the Shield icon, not "ON".

gigglearrows:
I think nordvpn is not naming its connection. It is called tun, but Ill look into it a bit closer. However, the tun0 interface means that you are connected using openvpn, so I think it is great to have it last in the list as a fallback as you suggested! im a bit busy this weekend, but ill test your changes

gigglearrows:
I added your suggestions!

scripts/nm-vpn
```diff
@@ -1,16 +1,16 @@
...
 
 CONNECTION=$(nmcli -t connection show --active | awk -F ':' '
-/tun0/{vpn="ON"} /vpn/{name=$1}
-END{if(vpn) printf("%s", name)}')
+/wg0|tun0/{vpn="ON"}; {if(/wireguard|vpn/) {vpn_name=$1} else if (/tun0/){tun_name=$1}}
+END{if ( (vpn) && (vpn_name)) {printf("%s", vpn_name)}  else if( (vpn) && (tun_name)) {printf("%s", tun_name)}}')
 
-LABEL_COLOR=${label_color:-$(xrescat i3xrocks.label.color)}
-VALUE_COLOR="${color:-$(xrescat i3xrocks.value.color)}"
+LABEL_COLOR=${vpn_label_color:-$(xrescat i3xrocks.vpn.label_color)}
```
leosunmo:
```suggestion
LABEL_COLOR=${vpn_label_color:-$(xrescat i3xrocks.label.color)}
```

scripts/nm-vpn
```diff
@@ -1,16 +1,16 @@
...
+VALUE_COLOR="${vpn_color:-$(xrescat i3xrocks.vpn.color)}"
```
leosunmo:
```suggestion
VALUE_COLOR="${vpn_color:-$(xrescat i3xrocks.value.color)}"
```

PR #12, opened 2019-11-24
Fixed the net icon issue

cheginit:
I added two function for font and label span to make the easier to read and fixed the net icon issue

### DIFF
--- a/scripts/nettraffic
+++ b/scripts/nettraffic
@@ -2,7 +2,7 @@
 
 # span for text
 fspan(){
-    echo "<span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\">${1}</span>"
+    echo "<span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\"> ${1}</span>"
 }
 
 # span for label/icon
--- a/scripts/nettraffic
+++ b/scripts/nettraffic
@@ -185,7 +185,7 @@ AN=`echo $AX/$FAC | bc -l | awk '{ printf("%4.1f", $1)}'`
 if [ "$RT" = "up" ]; then
     echo "$(lspan "${NIC}")$(fspan "$TN$SUF")$(lspan "${UP}")"
 elif [ "$RT" = "down" -o "$RT" = "dn" ]; then
-    echo "$(lspanj"${NIC}")$(fspan "$RN$SUF")$(lspan "${DN}")"
+    echo "$(lspan "${NIC}")$(fspan "$RN$SUF")$(lspan "${DN}")"
 elif [ "$RT" = "total" ]; then
     echo "$(lspan "${NIC}")$(fspan "$AN$SUF")"
 else
--- a/scripts/nettraffic
+++ b/scripts/nettraffic
@@ -1,27 +1,43 @@
 #!/bin/sh
 
+# span for font
+fspan(){
+    echo "<span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\"> ${1}</span>"
+}
+
+# span for label/icon
+lspan(){
+    echo "<span color=\"${LABEL_COLOR}\"> ${1}</span>"
+}
+
+# get font specifics from xresource file
 VALUE_COLOR=${color:-$(xrescat i3xrocks.value.color "#D8DEE9")}
 LABEL_COLOR=${label_color:-$(xrescat i3xrocks.label.color "#7B8394")}
-
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font "Source Code Pro Medium 13")}
 
+# read the dev file for extracting net usage
 GET="cat /proc/net/dev"
+
+# some default values for arguments"
 LABEL=true
 UPL=${label_icon:-$(xrescat i3xrocks.label.up ▴)}
 DNL=${label_icon:-$(xrescat i3xrocks.label.dn ▾)}
 ULEV=2
 
+# determine the net interface name
 IF=`ip route get 1.1.1.1 | grep -Po '(?<=dev\s)\w+' | cut -f1 -d ' '`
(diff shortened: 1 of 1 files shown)


scripts/nettraffic
```diff
@@ -1,27 +1,43 @@
 #!/bin/sh
 
+# span for font
```
kgilmer: I think this should say "span for text"

scripts/nettraffic
```diff
@@ -1,27 +1,43 @@
...
 VALUE_COLOR=${color:-$(xrescat i3xrocks.value.color "#D8DEE9")}
 LABEL_COLOR=${label_color:-$(xrescat i3xrocks.label.color "#7B8394")}
-
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font "Source Code Pro Medium 13")}
 
+# read the dev file for extracting net usage
 GET="cat /proc/net/dev"
+
+# some default values for arguments"
 LABEL=true
```
kgilmer: The "l" option also sets this to true.  Seems like either it should default to false or the flag should flip the value.

scripts/nettraffic
```diff
@@ -48,7 +62,6 @@ do
       echo -e "  -e  (0|1|2|3|4)\t Level of units displayed. (0 for none, 1 for K, 2 for KB, 3 for K/s,"
```
kgilmer: Not sure why, but the "-e" option is being printed to the console for me:
```
$ scripts/nettraffic -h
-e Usage: nettraffic [OPTION]
-e Options:
-e   -i  (%ifname%)      Specify which network interface to monitor. Required.
-e   -a  (up|dn|total)   Shows which data to display. Prints both if not used.
-e   -u  (K|KB|M|MB)     Units to show the data in. Automatic if not specified.
-e   -e  (0|1|2|3|4)     Level of units displayed. (0 for none, 1 for K, 2 for KB, 3 for K/s,
-e                       4 for KB/s). Default if 4.
-e   -l                  Choose whether or not to display the label (up or down).
-e   -t  (%integer%)     Set the interval for the data update. Default is 1.
-e   -v                  Prints version number.
-e   -h                  Prints this help.
For full help, visit http://github.com/ma-santoso/nettrafic/
```

I tried removing them from the echo commands and other than the tab characters ("\t") it's fine.  I suggest just removing the "-e" and tab characters.

scripts/nettraffic
```diff
@@ -1,27 +1,43 @@
 #!/bin/sh
 
+# span for font
+fspan(){
```
kgilmer: Good idea!  :+1:

scripts/nettraffic
kgilmer: Documentation says 4 is the default but `ULEV` is set to 2 above.

kgilmer: Also, typo in text "Default is 4."

scripts/nettraffic
```diff
@@ -1,27 +1,43 @@
...
+    echo "<span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\"> ${1}</span>"
+}
+
+# span for label/icon
+lspan(){
+    echo "<span color=\"${LABEL_COLOR}\"> ${1}</span>"
```
kgilmer: The spans for "icon fonts" should not use the space character as that maps to something other than space in some font character sets.  Here is what this version of nettraffic looks like with materialdesign instead of fontawesome:
image

but I change the expression to `echo "<span color=\"${LABEL_COLOR}\">${1}</span>"` there are no garbage characters.

scripts/nettraffic
```diff
@@ -1,27 +1,43 @@
...
+# read the dev file for extracting net usage
 GET="cat /proc/net/dev"
+
+# some default values for arguments"
 LABEL=true
 UPL=${label_icon:-$(xrescat i3xrocks.label.up ▴)}
 DNL=${label_icon:-$(xrescat i3xrocks.label.dn ▾)}
 ULEV=2
 
+# determine the net interface name
```
kgilmer: nice!

scripts/nettraffic
```diff
@@ -159,29 +176,18 @@ if [ "$UNIT" = "" ]; then
   fi
 fi
 
+# format net usage output with fixed width
 RN=`echo $RX/$FAC | bc -l | awk '{ printf("%4.1f", $1)}'`
 TN=`echo $TX/$FAC | bc -l | awk '{ printf("%4.1f", $1)}'`
 AN=`echo $AX/$FAC | bc -l | awk '{ printf("%4.1f", $1)}'`
 
+# output net usage using pango markup
```
kgilmer: So much more readable!

kgilmer:
I tested various network scenarios and wasn't able to get the character width to change, so looks like it's fixed!  I also agree with you that not having the leading zeros looks much better.  Thanks!

scripts/nettraffic
cheginit: Oh, I see. I was wondering about that. Great!

kgilmer:
I've tested your changes, everything looks good except one minor tweak.  I think we need a little more space between the network icon and the first metric.  Pictures are better than words:

image

cheginit:
> I've tested your changes, everything looks good except one minor tweak. I think we need a little more space between the network icon and the first metric. Pictures are better than words:
> 
> image

Yeah, agreed. This was the reason that I added space before icons but as you mentioned it doesn't work with some fonts like material. Any ideas?

kgilmer:
I believe the space should be prefixed to the text span, instead of the
icon span.

cheginit:
@kgilmer I added the space, please give it a try and let me knwo if it works.

kgilmer:
Perfect, well done @cheginit !  In fact a user asked for this feature today in Slack.

cheginit:
Great!

PR #11, opened 2019-11-13
Added script mediaplayer2 with pango markup and font awesome glyphs

Interesting-name:

PR #13, opened 2019-12-01
adding focused-window-name to show title of focused window

priestoferis:
This is I think my third pull request ever and the first one on a new function rather then just a tweak so I'm not sure if I did this right ...

Anyway, the script returns the title of the focused window. The short version is not really smart, but at least is short. It can also be used with command line arguments. I'm not sure this feature will be useful for anything, but I coded it with that first and then realized I need to use environment variables if I want a nice config in i3xrocks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ install:
 	$(INSTALL) -m0755 -D scripts/cpu_usage $(SCRIPT_TARGET)/cpu_usage
 	$(INSTALL) -m0755 -D scripts/disk $(SCRIPT_TARGET)/disk
 	$(INSTALL) -m0755 -D scripts/iface $(SCRIPT_TARGET)/iface
+	$(INSTALL) -m0755 -D scripts/focused_window_name $(SCRIPT_TARGET)/focused_window_name
 	$(INSTALL) -m0755 -D scripts/keyboard_layout $(SCRIPT_TARGET)/keyboard_layout
 	$(INSTALL) -m0755 -D scripts/keyindicator $(SCRIPT_TARGET)/keyindicator
 	$(INSTALL) -m0755 -D scripts/load_average $(SCRIPT_TARGET)/load_average
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -97,3 +97,10 @@ interval=10
 [time]
 interval=60
 
+# Show title of focused window
+[focused-window-name]
+label= 
+#interval=1
+#filter_list=evince libreoffice # uncomment to show only for these window instances
+#short_length=10 # uncomment to change maximum size of shortened text
+
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -98,8 +98,8 @@ interval=10
 interval=60
 
 # Show title of focused window
-[focused-window-name]
-label= 
+#[focused-window-name]
+#label=  # uncomment to show an icon
 #interval=1
 #filter_list=evince libreoffice # uncomment to show only for these window instances
 #short_length=10 # uncomment to change maximum size of shortened text
--- a/debian/copyright
+++ b/debian/copyright
@@ -35,6 +35,10 @@ Copyright: Copyright © 2014 Julien Bonjean <julien@bonjean.info>
            Copyright (C) 2014 Alexander Keller <github@nycroth.com>
 License: GPL-3+
 
+Files: scripts/focused-window-name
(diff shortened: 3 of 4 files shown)


i3xrocks.conf
```diff
@@ -97,3 +97,10 @@ interval=10
 [time]
 interval=60
 
+# Show title of focused window
+[focused-window-name]
```
kgilmer: I would like to keep this disabled by default until/if we determine it is important enough that all users get this blocket by default.

Makefile
```diff
@@ -15,6 +15,7 @@ install:
 	$(INSTALL) -m0755 -D scripts/cpu_usage $(SCRIPT_TARGET)/cpu_usage
 	$(INSTALL) -m0755 -D scripts/disk $(SCRIPT_TARGET)/disk
 	$(INSTALL) -m0755 -D scripts/iface $(SCRIPT_TARGET)/iface
+	$(INSTALL) -m0755 -D scripts/focused_window_name $(SCRIPT_TARGET)/focused_window_name
```
kgilmer: NIce!

i3xrocks.conf
priestoferis: Wow, github has a lot of features I didn't know about :) Anyway, I pushed a commit that adds a bit more text in comments, but comments out the entire section.

priestoferis:
I disabled the blocklet by default. 

Regarding the label, currently the icon is display by having the `label` section set in `i3xrocks.conf`. Does that have specific drawbacks?

Regarding the pango markup. The idea would be to read with `xrescat` the values set in the config like this:

```
```

and use these values to mark up? As far as I can see there is three of interest:
```
i3xrocks.value.color 
i3xrocks.label.color
i3xrocks.value.font
```

If that's the case I'll push a fix for pango.

priestoferis:
label is still added from the config but I added the formatting now it uses the same fonts as the other blocks

kgilmer:
In testing the script I get:
```
$ /usr/share/i3xrocks/focused-window-name
Traceback (most recent call last):
  File "/usr/share/i3xrocks/focused-window-name", line 30, in <module>
    from i3ipc import Connection, Event
ModuleNotFoundError: No module named 'i3ipc'
```

So, this dependency, i3ipc for python, will have to be expressed as a package dependency to ensure users have the necessary libraries installed before running the script.  I looked around and could not find a Debian package for this, so it will have to be packaged before this PR can be merged.  A similar situation to what's going on with https://github.com/regolith-linux/regolith-i3xrocks-config/pull/11.

In addition, since it's requires a specific dependency, it should probably be in a discrete package.

kgilmer:
Added i3ipc here: https://github.com/regolith-linux/speed-ricer/commit/bc57e8bf9cb5803415e1c7068efd314fdcc80f0f

kgilmer:
For reference: https://github.com/altdesktop/i3ipc-python/issues/139

priestoferis:
Oh yes, sorry about that, I had it on my computer for some time and forgot about it ... I'm not sure how to proceed with this. You might think about including i3ipc by default as it might be used to solve nicely the "have a keybind to open the next workspace" problem as well.

kgilmer:
I think the Debian packaging format will allow me to specify a second binary package in which we can ship this blocklet with the declared dependency of i3ipc-python.  So, I'll work on this issue in the packaging code.

kgilmer:
I will test the blocklet after installing my packaging of i3ipc-python, here: https://launchpad.net/~kgilmer/+archive/ubuntu/speed-ricer/+sourcepub/10773128/+listing-archive-extra

priestoferis:
This is a cosmetic update, LibreOffice puts the name of the app into the title i.e `Example.doc - LibreOffice Writer` which just takes up space since you can already see which window is focused anyway.

kgilmer:
Successfully tested after installing i3ipc-python and python3-xlib:

```
$ ./focused-window-name
<span font_desc="Source Code Pro Medium 13" color = "#D8DEE9" > st </span>
<span font_desc="Source Code Pro Medium 13" color = "#D8DEE9" > st </span>
```

I did not add a required dependency into the i3ipc-python package.  I've building [that change](https://github.com/regolith-linux/speed-ricer/compare/bc57e8bf9cb5...8d34efe3acaa) now, but am ready to merge this in.  Thanks @priestoferis for your awesome new blocket!

priestoferis:
cool thanks

PR #14, opened 2019-12-04
Adding a script temperature2 that follows the selected color scheme

bdzzz:
Adding a new temperature script that uses the _value, warning_ and _critical_ colors from the color scheme.

### DIFF
--- a/scripts/temperature2
+++ b/scripts/temperature2
@@ -4,12 +4,12 @@
 
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font "Source Code Pro Medium 13")}
 
-TEMP=$(sensors -u | grep -Po 'temp1_input: \K\d*')
+TEMP=$(sensors -u | grep -Po 'temp1_input: \K\d*' | head -n 1)
 
-if [ "$TEMP" -gt 90 ]
+if [[ ${TEMP} -gt 90 ]]
 then
   COLOR=${critical_color:-$(xrescat i3xrocks.critical.color "#BF616A")}
-elif [ "$TEMP" -gt 70 ]
+elif [[ ${TEMP} -gt 70 ]]
 then
   COLOR=${warning:-$(xrescat i3xrocks.warning "#EBCB8B")}
 else
--- a/scripts/temperature2
+++ b/scripts/temperature2
@@ -3,6 +3,8 @@
 # Depends on lm-sensors (https://packages.ubuntu.com/bionic/lm-sensors)
 
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font "Source Code Pro Medium 13")}
+LABEL_ICON=${label_icon:-$(xrescat i3xrocks.label.thermometer )}
+LABEL_COLOR=${label_color:-$(xrescat i3xrocks.label.color "#7B8394")}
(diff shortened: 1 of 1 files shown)


scripts/temperature2
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Depends on lm-sensors (https://packages.ubuntu.com/bionic/lm-sensors)
+
+VALUE_FONT=${font:-$(xrescat i3xrocks.value.font "Source Code Pro Medium 13")}
+
+TEMP=$(sensors -u | grep -Po 'temp1_input: \K\d*')
```
kgilmer: This adds a dependency on lm-sensors.  Happily it looks like this package is well supported in Ubuntu: https://launchpad.net/ubuntu/+source/lm-sensors/1:3.5.0-3ubuntu1

scripts/temperature2
```diff
@@ -0,0 +1,19 @@
...
+
+if [ "$TEMP" -gt 90 ]
```
kgilmer: The conditional expression isn't parsing for me.  Testing:

```
$ ./temperature2
./temperature2: line 9: [: 37
27: integer expression expected
./temperature2: line 12: [: 37
27: integer expression expected
<span font_desc="Source Code Pro Medium 13" color="#D8DEE9">37
27℃</span>
```

My version of bash: 
```
$ bash --version
bash --version
GNU bash, version 5.0.3(1)-release (x86_64-pc-linux-gnu)
```

bdzzz: I managed to reproduce and fix it. 
I also forced the command to only look at the first temperature it finds. This follows the logic of the existing [temperature](https://github.com/regolith-linux/regolith-i3xrocks-config/blob/master/scripts/temperature#L48-L53) script.

kgilmer:
Looks good @bdzzz !  One last thing is that the Regolith blockets have an "icon" as a prefix that is intended to provide a hint to the user as to the meaning of the metric.  The default icons are provided by fontawesome in R1.2.  I'm considering moving to materialdesign for R1.3.  For, fontawesome, how about https://fontawesome.com/icons/thermometer-full?style=solid and for material design how about  https://materialdesignicons.com/icon/thermometer ?

This change requires adding a key value pair to the Xresources defined in the gh repo [regolith-styles](https://github.com/regolith-linux/regolith-styles).  I'm happy to add the value and update the script but leave it up to you if you'd like to make the change yourself.  :smile:

kgilmer:
In particular this file is where the key value pair is defined: https://github.com/regolith-linux/regolith-styles/blob/master/Xresources/i3xrocks

And in these files the specific characters are set which get included based on the user's preferences:

https://github.com/regolith-linux/regolith-styles/blob/master/Xresources/typeface-sourcecodepro
https://github.com/regolith-linux/regolith-styles/blob/master/Xresources/typeface-ubuntu
https://github.com/regolith-linux/regolith-styles-typeface-materialdesign/blob/master/typeface-sourcecodepro-materialdesign

bdzzz:
The icons look really cool :open_mouth:. I'll add them and submit the change.

kgilmer:
Beautiful!

PR #15, opened 2019-12-04
display date according to locale

ploum:
and not only according to US rules

### DIFF
--- a/i3xrocks.conf
+++ b/i3xrocks.conf
@@ -96,4 +96,4 @@ interval=10
 # Date Time
 [time]
 interval=60
-
+command=date "+%x %X"


i3xrocks.conf
```diff
@@ -96,4 +96,4 @@ interval=10
 # Date Time
 [time]
 interval=60
-
+command=date "+%x %X"
```
kgilmer: I had no idea the date command handled locale formats!  I looked in gnome for something but only found 12/24 hour format specification.  Thanks for submitting this PR @ploum .  I think the format change should go here instead: https://github.com/regolith-linux/regolith-styles/blob/master/Xresources/i3xrocks#L8

ploum:
thanks, I submitted another PR.

PR #22, opened 2020-02-12
Resolve negative speed issue

cheginit:
I tried to fix the negative speed issue but I have no way of testing it! I also checked the script with [shellcheck](https://www.shellcheck.net/) and fixed the issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .pc
 debian/files
+.nvimlog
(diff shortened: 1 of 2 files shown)


kgilmer:
Very clean, thank you for the fix @cheginit !

jc00ke:
I'll test when once it's updated. Thanks for the fix!

PR #23, opened 2020-02-15
Update battery3

baldrailers:
Add better icon/label for charging state. This should address [#235](https://github.com/regolith-linux/regolith-desktop/issues/235)

### DIFF
--- a/scripts/battery3
+++ b/scripts/battery3
@@ -51,9 +51,9 @@ if [ -z "$CHARGE_STATE" ]; then
     elif [ "$CHARGE_STATE" == "discharging" ]; then
     CHARGE_ICON=$(xrescat i3xrocks.label.dn D)
     elif [ "$CHARGE_STATE" == "charging" ]; then
-    CHARGE_ICON=$(xrescat i3xrocks.label.batterycharging C)
+    CHARGE_ICON=$(xrescat i3xrocks.label.battery_charging C)
     elif [ "$CHARGE_STATE" == "fully-charged" ]; then
-    CHARGE_ICON=$(xrescat i3xrocks.label.batteryac F)
+    CHARGE_ICON=$(xrescat i3xrocks.label.battery_ac F)
 else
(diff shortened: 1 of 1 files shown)


scripts/battery3
```diff
@@ -51,7 +51,9 @@ if [ -z "$CHARGE_STATE" ]; then
     elif [ "$CHARGE_STATE" == "discharging" ]; then
     CHARGE_ICON=$(xrescat i3xrocks.label.dn D)
```
kgilmer: If we're going away from the up = charge, down = discharge semantics, we'll need to replace discharging w/ something else to remain consistent.

scripts/battery3
```diff
@@ -51,7 +51,9 @@ if [ -z "$CHARGE_STATE" ]; then
...
     elif [ "$CHARGE_STATE" == "charging" ]; then
-    CHARGE_ICON=$(xrescat i3xrocks.label.up C)
+    CHARGE_ICON=$(xrescat i3xrocks.label.batterycharging C)
```
kgilmer: These Xresource keys have underscores.  See https://github.com/regolith-linux/regolith-styles/blob/master/cahuella/typeface#L18-L19

scripts/battery3
baldrailers: I agree. Right now the change will address the `charging,full-charged` status, and I can't think of any replacement label for `discharging`. So we can put this on hold until we get something that makes more sense than the current label.

scripts/battery3
baldrailers: Ah, yeah the labels I have was from 1.2 release that's why there were no underscore. I'll update the PR. Thanks!

scripts/battery3
kgilmer: Do you mean to revert the change on line 54 or are you looking for a replacement for line 52?

baldrailers:
Closing this as the version of the scripts I'm running is base from 1.2

PR #26, opened 2020-03-16
Added support for tap0

ivomarino:
This simple modification adds support for tap devices, may be useful also to other users.


kgilmer:
Thank you @ivomarino LGTM!

PR #27, opened 2020-03-25
Alternative 'volume' script

gdeflaux:
Based on the 'volume' script.

Fixed bug that made the mute icon not to show even if muted.
Simplified the script by creating functions for each step (improved readability in my opinion).
Split 'step' and 'unit' in 2 separate arguments. Defaults to step=5 and unit=%.

### DIFF
--- a/scripts/volume2
+++ b/scripts/volume2
@@ -90,7 +90,7 @@ case $BLOCK_BUTTON in
 esac
 
 ICON_ON=${icon_on:-$(xrescat i3xrocks.label.sound S)}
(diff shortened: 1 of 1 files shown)


kgilmer:
Works great for me @gdeflaux , thanks!  Since you've fixed a bug in the original volume switch, any concerns for me replacing the original with this, instead of adding a second script?

gdeflaux:
Sure go ahead!

kgilmer:
Manually merged the change, thanks again @gdeflaux !

PR #28, opened 2020-03-27
Modify pulse audio mixer detection

dpint:
PulseAudio isn't detected on Ubuntu 18.04, which in turn means that default mixer gets used. When default mixer is used, it is not possible to unmute audio with amixer command. This problem is reported here: https://bugs.launchpad.net/ubuntu/+source/pulseaudio/+bug/878986

This pull request introduces changes from an upstream repository regarding PulseAudio detection:
https://github.com/vivien/i3blocks-contrib/blob/master/volume/volume


kgilmer:
Thank you for the detailed description of the problem and fix @dpint !  :smile: .

When testing your change I am getting this error when running the script from the command line:
```
$ /usr/share/i3xrocks/volume2
/usr/share/i3xrocks/volume2: line 72: [: too many arguments
/usr/share/i3xrocks/volume2: line 82: [: too many arguments
```

I add the brackets and resolve the issue but the unmute does not work (original issue).  I may have messed something up, although it's not apparent to me.  Here is my edit of your changes: 
```
#!/bin/bash
# Copyright (C) 2014 Julien Bonjean <julien@bonjean.info>
# Copyright (C) 2014 Alexander Keller <github@nycroth.com>

# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.

# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.

# You should have received a copy of the GNU General Public License
# along with this program.  If not, see <http://www.gnu.org/licenses/>.

#------------------------------------------------------------------------

# The second parameter overrides the mixer selection
# For PulseAudio users, use "pulse"
# For Jack/Jack2 users, use "jackplug"
# For ALSA users, you may use "default" for your primary card
# or you may use hw:# where # is the number of the card desired
MIXER="default"
if [ command -v pulseaudio >/dev/null 2>&1 && pulseaudio --check ]; then
    # pulseaudio is running, but not all installations use "pulse"
    if [ amixer -D pulse info >/dev/null 2>&1 ]; then
        MIXER="pulse"
    fi
fi
[ -n "$(lsmod | grep jack)" ] && MIXER="jackplug"
MIXER="${2:-$MIXER}"

# The instance option sets the control to report and configure
# This defaults to the first control of your selected mixer
# For a list of the available, use `amixer -D $Your_Mixer scontrols`
SCONTROL="${BLOCK_INSTANCE:-$(amixer -D $MIXER scontrols |
        sed -n "s/Simple mixer control '\([A-Za-z ]*\)',0/\1/p" |
        head -n1
)}"

# The first parameter sets the step to change the volume by (and units to display)
# This may be in in % or dB (eg. 5% or 3dB)
STEP="${1:-5%}"

#------------------------------------------------------------------------

capability() { # Return "Capture" if the device is a capture device
    amixer -D $MIXER get $SCONTROL |
    sed -n "s/  Capabilities:.*cvolume.*/Capture/p"
}

volume() {
    amixer -D $MIXER get $SCONTROL $(capability)
}

format() {
    perl_filter='if (/.*\[(\d+%)\] (\[(-?\d+.\d+dB)\] )?\[(on|off)\]/)'
    perl_filter+='{CORE::say $4 eq "off" ? "$MUTE_ICON" : "'
    # If dB was selected, print that instead
    perl_filter+=$([[ $STEP = *dB ]] && echo '$3' || echo '$1')
    perl_filter+='"; exit}'
    perl -ne "$perl_filter"
}

#------------------------------------------------------------------------

case $BLOCK_BUTTON in
    3) amixer -q -D $MIXER sset $SCONTROL $(capability) toggle ;;  # right click, mute/unmute
    4) amixer -q -D $MIXER sset $SCONTROL $(capability) ${STEP}+ unmute ;; # scroll up, increase
    5) amixer -q -D $MIXER sset $SCONTROL $(capability) ${STEP}- unmute ;; # scroll down, decrease
esac

LABEL_ICON=${label_icon:-$(xrescat i3xrocks.label.sound S)}
MUTE_ICON=${mute_icon:-$(xrescat i3xrocks.label.mute M)}

VALUE=$(volume | format)

echo "<span color=\"${LABEL_COLOR}\">${LABEL_ICON}</span><span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\"> $VALUE</span>"
```
My testing steps:
1. Add your changes to `/usr/share/i3xrocks/volume2`
2. play some music
3. mute / un mute
Expected: un mute works
Actual: audio stays muted

Any suggestions?

kgilmer:
Also, another user has made some fixes to the original script that I just pushed in this commit: https://github.com/regolith-linux/regolith-i3xrocks-config/commit/0b15c666384bab3edcb0ddefc149653e5cc7cd66

dpint:
Hmm, interestingly both scripts execute and work without any errors for me. Does sound mute/unmute work if you execute "amixer -D pulse sset Master toggle" command on your device? I think you may need to replace "Master" with a different control. You can see controls on your device with "amixer -D pulse scontrols" command.

dpint:
I installed Ubuntu 20.20 today and I noticed I get the same error as you and that the fix wasn't working. Debugging it I discovered that the issue lies in multiple sound outputs that are returned by "amixer -D pulse get Master" command (as seen below), due to which "on\non" is returned as a status value, whereas we only expect "(on|off)". In my fix I simply grab the first line, but this may not be the perfect way to handle this issue as it doesn't handle the case where only one of the outputs may be enabled.

Simple mixer control 'Master',0
  Capabilities: pvolume pswitch pswitch-joined
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 65536
  Mono:
  Front Left: Playback 48357 [74%] [on]
  Front Right: Playback 48357 [74%] [on]

Anyway, it seems that something was different on my previous install regarding PulseAudio as I didn't have the same issue on my previous install.

I also noticed someone else submitted pull request for the PulseAudio detection. It is up to you which one you want to merge.

kgilmer:
Hi @dpint sorry, I'd forgotten about this PR when merging the other one.  Let's keep this open for now and see how testing goes with @agoose77 's changes.

agoose77:
Thanks @kgilmer, i missed this too. I like the fact that this PR checks the output of a mixer before assigning MIXER to pulse.

kgilmer:
@agoose77 can this PR be closed?

kgilmer:
Closing

PR #30, opened 2020-04-09
disk usage block

lejeunel:
This shows the disk usage for a given mount point

The block declaration follows:
`
[disk]
command=~/.config/regolith/i3blocks/disk
LABEL=home
MNT=/home
interval=60
`

### DIFF
--- a/scripts/disk
+++ b/scripts/disk
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


kgilmer:
Beautiful, thank you @lejeunel !  What shall I put for your name in the attribution?

lejeunel:
With great pleasure.
You may write "Laurent Lejeune".

PS: I am currently trying to integrate Xresources configuration to rofication. See [https://github.com/lejeunel/regolith-rofication]. The goal being that one can provide custom keybindings.
The current solution parses a provided config directly from Python at each execution.
Let me know if you have any recommendation/critic.

Best,

L.

PR #32, opened 2020-04-14
Change awk logic to detect right connection name.

dennis-reiter:
I made a few changes to the part of the script detecting and choosing the right vpn connection name.
Here is a working example from nmcli:
```MY_VPN:ID1:vpn:wlo1
MY_WIFI:ID2:802-11-wireless:wlo1
vpn0:ID3:tun:vpn0
docker0:ID4:bridge:docker0
```
With the previous version, I got none. The new one finds and selects `MY_VPN`. But I am not sure about other test cases, especially with wireguard.

### DIFF
--- a/scripts/nm-vpn
+++ b/scripts/nm-vpn
@@ -1,9 +1,15 @@
 #!/bin/bash
 
 CONNECTION=$(LC_ALL=C nmcli -t connection show --active | awk -F ':' '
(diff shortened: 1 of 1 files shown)


kgilmer:
Thanks @dennis-reiter , looks good to me, although I have no way of testing.  I've posted in a [long running bug](https://github.com/regolith-linux/regolith-desktop/issues/64) to see if anyone is able to test those other code paths.  If nothing comes of it I'll merge the CR but would be better if we could get more testing.

vanboom:
Happy to help test this, but unfortunately this is over my head.

kgilmer:
Let's move forward!

PR #33, opened 2020-04-26
No limit for weather block length

gdeflaux:
Hi,

I suggest removing the test on the length of the block string. Depending on the format set by the user, valid strings can be rejected. Ex: a format that displays the name of a city that is more than 10 characters long (San Francisco), or a format that displays wind or other parameters.

Printing whatever is returned by the weather service is a good way to debug. Printing nothing can be confusing.

### DIFF
--- a/scripts/weather
+++ b/scripts/weather
@@ -13,11 +13,7 @@ VALUE_WEATHER_LOCATION=${weather_location:-$(xrescat i3xrocks.weather.location "
 WEATHER_DATA=`curl -sS wttr.in/${VALUE_WEATHER_LOCATION}?format="${VALUE_WEATHER_FORMAT}"`
 WEATHER_DATA_LEN=${#WEATHER_DATA}
 
-if [ $WEATHER_DATA_LEN -gt 10 ]; then
-    echo ""
-else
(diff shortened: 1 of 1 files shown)


kgilmer:
Thanks @gdeflaux , the reason I added this length check in the beginning is that if there is some kind of network error, that error text is passed back to i3xrocks, and because it's very long ends up filling the entire bar.  Is there a way we can deal w/ this without a length restriction?

gdeflaux:
We could  use the return status of the HTTP request. Anything other than 200 would be considered as an error.

Example below. In case of error a strike through cloud icon is displayed, but we can decide to replace by anything else. I would be in favor of displaying something to inform the user something is wrong.

```sh
WEATHER_DATA_FILE=/tmp/i3xrocks_weather_data
RESPONSE_CODE=`curl -o ${WEATHER_DATA_FILE} -sS -w "%{http_code}" wttr.in/${VALUE_WEATHER_LOCATION}?format="${VALUE_WEATHER_FORMAT}"`

if [ $RESPONSE_CODE -ne "200" ]; then
    echo ""
else
    WEATHER_DATA=`cat ${WEATHER_DATA_FILE}`
    echo "<span color=\"${LABEL_COLOR}\">$LABEL_ICON</span><span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\">${WEATHER_DATA}</span>"
fi
```

kgilmer:
+1.  Great ideas.

PR #34, opened 2020-04-27
Add quoting to fix error when variable is empty

agoose77:


kgilmer:
Thank you @agoose77 !

PR #36, opened 2020-04-27
Take only first volume and status from amixer

agoose77:
My system reports two audio values, i.e.:
`amixer -D pulse get Master`

```
Simple mixer control 'Master',0
  Capabilities: pvolume pswitch pswitch-joined
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 65536
  Mono:
  Front Left: Playback 15309 [23%] [on]
  Front Right: Playback 15309 [23%] [on]
```
This change proposes to just take the first. I don't know if this is good practice or not, would appreciate any input.

### DIFF
--- a/scripts/volume
+++ b/scripts/volume
@@ -59,7 +59,7 @@ mixer_info() {
 
 # Returns "on" if not mute, "off" if mute
 status() {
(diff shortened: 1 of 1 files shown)


kgilmer:
I don't have any suggestions at the moment.  Taking the first input seems like a reasonable improvement.

PR #37, opened 2020-04-27
Determine pulseaudio using `pactl list`

agoose77:
Replaces lsmod check with `pactl list` return code. On my distro, pulse doesn't appear in lsmod. Similarly, just check the retcode for grep rather than doing a string substitution

Again, this might not be the best fix, e.g. what if pulse audio is installed, but lists no sinks/devices etc.

### DIFF
--- a/scripts/volume
+++ b/scripts/volume
@@ -36,8 +36,8 @@ UNIT="${2:-%}"
 # For ALSA users, you may use "default" for your primary card
 # or you may use hw:# where # is the number of the card desired
 MIXER="default"
-[ -n "$(lsmod | grep pulse)" ] && MIXER="pulse"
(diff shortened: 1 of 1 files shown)


kgilmer:
Looks good, I tested it locally and seems fine to me.

kgilmer:
@agoose77 your changes are ready for testing on Focal in the `unstable` PPA.  LMK if you need a build on a different version of Ubuntu.

https://launchpad.net/~regolith-linux/+archive/ubuntu/unstable/+sourcepub/11249629/+listing-archive-extra

You can either download the deb and install it w/ `dpkg` or add `ppa:regolith-linux/unstable` to your PPAs.  I suggest the former.

PR #38, opened 2020-04-27
i3xrocks battery blocklet should not emit the charging span when fully charged

farsil:
As for the subject, this PR addresses a minor graphical issue where the `battery` i3xrocks blocklet displays a larger gap when the battery is fully charged.

This happens because the "charging" span in the script is emitted in any case, even if the battery is not charging or discharging.

With this PR, the `battery` blocklet only emits the "charging" span if the battery is actually charging or discharging.

### DIFF
--- a/scripts/battery
+++ b/scripts/battery
@@ -29,16 +29,16 @@ CHARGE_STATE=$(get_battery_status)
 
 if [ -z "$BATT_PERCENT" ]; then
     exit 0
-    elif [ "$BATT_PERCENT" -ge 0 ] && [ "$BATT_PERCENT" -le 20 ]; then
+elif [ "$BATT_PERCENT" -ge 0 ] && [ "$BATT_PERCENT" -le 20 ]; then
     LABEL_ICON=$(xrescat i3xrocks.label.battery0 E)
     VALUE_COLOR=$(xrescat i3xrocks.critical.color red)
-    elif [ "$BATT_PERCENT" -ge 20 ] && [ "$BATT_PERCENT" -le 50 ]; then
+elif [ "$BATT_PERCENT" -ge 20 ] && [ "$BATT_PERCENT" -le 50 ]; then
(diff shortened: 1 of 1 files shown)


kgilmer:
So the reason for the `CHARGE_ICON=$(echo -e '\u2003')` is to prevent the bar from resizing when textual values change.  It looks bad, and due to the nature of i3bar/i3blocks, what's passed is just text.  So there is no way of positioning a block at a specific location.  If the battery block suddenly takes an extra character, then all blocks to the left move one character over.

Otherwise, I like all the cleanup, and I'd like to take at least that.  If you find another solution to the positioning issue I'm happy to take it.

farsil:
Well, as an user, I would rather have all the text shift a bit to the left when my laptop is charging, than have a permanently on blank space messing with the gaps. 

Charging your laptop is an activity that takes time, so you only experience the blocklet shift only once in a while. It happens much more frequently with the memory/volume/network blocklet, at least with the way I use my laptop.

But if you still prefer to keep a blank space and keep the cleanup I can do that of course :)

kgilmer:
How about we count in base 16 so we can avoid 3 digits?  :smile_cat:   I understand your reasoning and while it's very subjective, I really dislike having the bar indicators moving around.  It gives me a sense of instability.  An earlier version of the battery script worked as you suggest, and while it's true that the transition from 100 <-> 99 is infrequent, it tends to happen to me a few minutes after unplugging the computer to go sit on the couch or something, so I ended up noticing it often.  So, I would like to keep the digit behavior as is.  If you'd like to implement an Xres override so that the behavior can be configured by the user I would be fine with that.  Your call.  Regardless, your other cleanup of this script is great and would like to include it as mentioned.

farsil:
That's fine, I was merely stating my opinion. I will provide the cleanup script after work. I have no problems keeping my patched version of the script for myself, after all i3 is meant to be customized :)

PR #39, opened 2020-04-27
Add space between text and icon

RoobieBoobieee:
This change adds a space between the text and the icon behind.

### DIFF
--- a/scripts/net-traffic
+++ b/scripts/net-traffic
@@ -177,13 +177,13 @@ done
 
 # output net usage using pango markup
 if [ "$RT" = "up" ]; then
-    echo "$(lspan "${NIC}")$(fspan "$TN$SUF")$(lspan "${UP}")"
+    echo "$(lspan "${NIC}")$(fspan "$TN$SUF") $(lspan "${UP}")"
     elif [ "$RT" = "down" ] || [ "$RT" = "dn" ]; then
-    echo "$(lspan "${NIC}")$(fspan "$RN$SUF")$(lspan "${DN}")"
+    echo "$(lspan "${NIC}")$(fspan "$RN$SUF") $(lspan "${DN}")"
     elif [ "$RT" = "total" ]; then
     echo "$(lspan "${NIC}")$(fspan "$AN$SUF")"
 else
-    echo "$(lspan "${NIC}")$(fspan "$RN$SUF")$(lspan "${DN}")$(fspan "$TN$SUF")$(lspan "${UP}")"
+    echo "$(lspan "${NIC}")$(fspan "$RN$SUF") $(lspan "${DN}")$(fspan "$TN$SUF") $(lspan "${UP}")"
 fi
 
 if [ ! -z "$button" ]; then


kgilmer:
Hi @RoobieBoobieee , can you provide a screenshot as to how this change is better than what's there now?

RoobieBoobieee:
Here you go: 
Before:
20200428-114013_before
After:
20200428-114119_after

The space prevents the icon from sticking against the text.
Note: I use custom icons from FontAwesome

kgilmer:
It's clear to me know, thanks for providing the screenshots.  I agree that it looks better with your custom icons, but it regresses the default in my opinion.  In fact, an earlier version of this script had the spaces.  I spent a little time trying to figure out another approach to the issue that would work for you and the default.  I tried simply specifying a character which included a space, like: 

```bash
$ echo "i3xrocks.label.up:  " >> ~/.config/regolith/Xresources
```

But unfortunately all empty space characters render as a "generic unicode" character.  I could not find a unicode value or otherwise that would produce the desired result.  This issue may be related: https://github.com/FortAwesome/Font-Awesome/issues/1606

I thank you for your time to put together this contribution but cannot merge it.  If you find a way of preventing the spacing to occur w/ the default characters then I would be happy to make the change.

buck10:
Hi Ken, I was digging through xrescat when I saw pull request, and it looks like you can get spaces into the strings in Xresources by escaping the space with a backslash, like:

`$ echo "i3xrocks.label.up:  \ " >> ~/.config/regolith/Xresources`

if you run xrescat i3xrock.label.up you should see the space printed before the icon. You could do the same after the icon if you need a space there as well. It looks to me like you need to escape all spaces if you want xrescat to output them exactly as you typed. Anyway, perhaps this could be a solution for this problem.

kgilmer:
Thanks for the suggestion @buck10 !  The issue is that the space character maps to not-a-space in the icon fonts.  Generally this is dealt with via Pango `<span>` tags to denote either a normal font (spaces are OK) and the icon font.  However at this level, the style templates are assuming that whatever value is defined for, say, `i3xrocks.label.up` is encoded in the icon font.  

If there were a space-like character that was available in both fontawesome and material design web icon font, then this would be a solution.

buck10:
Interesting. That is not what I am seeing when I insert escaped spaces after the icons in the bar.

kgilmer:
I may have been confusing another issue.  I tried this and was able to see the spacing properly:

```bash
echo "i3xrocks.label.up: \ " >> ~/.config/regolith/Xresources
```

@RoobieBoobieee can you check if this fixes your issue?

buck10:
I checked the use of the escape sequence more and it doesn't work properly if you want spaces after the last character. xrescat seems to want to ignore the space and then escape the linefeed, thereby pulling the next line in the Xresources file into the returned result.

I think a better approach would be to have a wrapper around the call to xrescat to allow us to use double-quotes around the values in the Xresources file that would then be stripped from the result when read back, which is more how set_from_resource work.

I wrote the following function that does this:

```
#!/usr/bin/env bash

XRESCAT () {                        
    local result;

    if [ ${#} -lt 1 ]; then
        #invalid usage, return nothing
        result=""
    elif [ ${#} -eq 1 ]; then
        # no default supplied, call xrescat with empty value
        result=`xrescat "${1}" ""`
    else
        result=`xrescat "${1}" "${2}"`
    fi

    # strip leading and trailing " if both are present and suppress the line feed
    echo -n "${result}" | sed 's/^\"\(.*\)\"$/\1/'
}
```
This could be put somewhere that all scripts can then source if they want so that any calls to xrescat could be replaced with XRESCAT. If there are no quotes at the start and end of the returned string, then the result returned from xrescat is not modified.

Not sure where the best place to put this file would be, but I suppose /usr/share/i3xrocks or /etc/regolith/i3xrocks would be options.

Let me know what you think.

RoobieBoobieee:
@kgilmer That works for me, thanks!

buck10:
@kgilmer, I have a patch for xrescat that does what my bash script above does, but in C. So it would mean no changes are needed to any of the i3xrocks scripts, but would let people put additional spaces before and after their resource strings if desired. I can issue a pull request if you like. It is currently in my fork of xrescat.

kgilmer:
Sounds interesting, yes please provide your PR for `xrescat` @buck10 !

buck10:
Sent the PR over.

kgilmer:
Closing as we took the xrescat PR.

PR #40, opened 2020-04-28
Patch to allow a better display of the block in case of error.

gdeflaux:


kgilmer:
Is there a way we can avoid writing to `/tmp`?  Looking at the curl man page, it looks like it returns error codes when things go wrong, and I've verified:

```bash
kgilmer@x280:~$ curl https://thisdoestd3.com
curl: (7) Failed to connect to thisdoestd3.com port 443: Connection refused
kgilmer@x280:~$ echo $?
7
```

So what about something like:

```
$ echo $(curl https://thisdoesntexist3242342.com || echo "boo" )
```

gdeflaux:
Good point. I've update the script as you suggested.

kgilmer:
Fantastic, thank you!

PR #41, opened 2020-04-28
Improve Geo localization to provide weather information

nachovizzo:
- Adds libxml2-utils and curl dependencies
- The default location service used for https://wttr.in
  is not accurate enough. In my case the error is about
  400km(600 miles)
- The i3xrocks.weather.location could still be used, so
  this doesn't break any current installation. In my
  case, I share the same configuration across multiple
  machines in different parts of the world. So it's
  better to pick the localization automatically.


cheginit:
Your solution doesn't work for my location. Maybe you can give this a try:

```bash
deg="F"  # or C for Celsius
ip="$( dig @resolver1.opendns.com ANY myip.opendns.com +short )"
city="$( curl -s https://ipapi.co/$ip/city/ )"
code="$( curl -s https://ipapi.co/$ip/region_code/ )"
loc="$( sed -e "s/ /%20/g" <<< "$city, $code" )"
url="https://thisdavej.azurewebsites.net/api/weather/current?loc=$loc&deg=$deg"
temp="$( curl -sS "$url" | grep -Po '"temperature":.*?[^\\]",' | cut -d "\"" -f 4 )"
```

This gives you the temperature. See if this works for your location as well.

nachovizzo:
Sorry for the late reply, sadly that doesn't work for me in terms of weather information. But it does give me the proper location.

Although it looks a bit more bloated than my solution, that being said, I have no complaints.

My original idea was to leave the script as it is(quite simple indeed), but improve the geolocalization default service, but if this doesn't work for you.... I guess that there is no global solution

cheginit:
If the location works for you then you can find the city using `ipapi` service and the weather from `wttr.in`:
```bash
ip="$( dig @resolver1.opendns.com ANY myip.opendns.com +short )"
city="$( curl -s https://ipapi.co/$ip/city/ )"
weather="$( curl -s "wttr.in/$city?format=3" )"
```

nachovizzo:
Works like a charm!

kgilmer:
We need a rebase...

kgilmer:
There is also this... https://github.com/regolith-linux/regolith-desktop/issues/372  Not sure if it's easy enough to integrate this change request or should be done separately...

nachovizzo:
@kgilmer Why rebase? 

I guess the ability to pick C/F is not strictly related to this PR.

kgilmer:
Because that what GitHub says regarding merging it:

```
This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.
```

C/F: sounds good

nachovizzo:
image

I will merge anyway

kgilmer:
I am confused!

image

nachovizzo:
I have no clue then! Github bug maybe? I can't merge obviously because I don't have write access.

kgilmer:
There are three ways of merging, the button I think was defaulted to my last action which wasn't possible due to commits.  Changing to a different merge type solved it.  My bad!

PR #42, opened 2020-04-28
fixed width output

cheginit:
This PR fixes the width of the output field to 25 characters. The original script depending on the song title and artists sometimes could take up the whole bar since there was no limit! This will takes care of it. If the output exceeds 25 characters it will add 3 dots to show that the text is cut off

image

image

### DIFF
--- a/scripts/media-player
+++ b/scripts/media-player
@@ -4,11 +4,12 @@
 STATUS=$(playerctl status 2> /dev/null)
 
 # use given values or fall back to parameters from xresources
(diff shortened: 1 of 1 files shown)


cheginit:
I made a [PR](https://github.com/regolith-linux/regolith-styles/pull/18) in `i3xrocks-confis` repo required for this PR to work.

kgilmer:
Beautiful, thanks @cheginit

PR #44, opened 2020-04-30
Fix focused-window-name plugin

rtfpessoa:
Windows names were wrongly including the string `uncomment to show an icon`


kgilmer:
Thanks for the fix @rtfpessoa !

rtfpessoa:
@kgilmer thank you for your work in regolith 🚀

PR #45, opened 2020-05-02
Udpated gsetting parameter lookup

gdeflaux:
The `sources` parameters is what works for me. `mru-sources` parameter is empty causing the block to display non-sense.

Screenshot from 2020-05-02 17-38-06


kgilmer:
Thanks @gdeflaux :smile: .  Verified same is true for bionic, eoan, and focal.

PR #47, opened 2020-05-04
Updates to the volume blocklet script

farsil:
This relatively localized PR actually changes many things in the volume blocklet. I do not expect all of the changes to be included upstream, so I have separated the changes in different commits, in order to make it easier to pick what to include, in a way.

### Changes
- `STEP`, `UNIT` and `MIXER` are now searched in `XResources` definitions before falling back to the default values. The three relevant definitions are `i3xrocks.volume.step`, `i3xrocks.volume.unit` and `i3xrocks.volume.mixer`.
- The volume level is now displayed even if the control is muted. There are two reasons for this:
  - The blocklet size stays consistent when muting/unmuting.
  - It can be useful to know the volume level before unmuting.
- The volume level is displayed in warning color when muted, to better highlight the reason the sound is not working,
- If the volume is changed while muted, the mixer is no longer automatically unmuted. This can be useful when you want to change the output volume before unmuting (maybe because it was too loud).
- Consistent function names and whitespace.

### DIFF
--- a/scripts/volume
+++ b/scripts/volume
@@ -72,13 +72,8 @@ icon() {
 }
 
 value() {
-    STATUS=$(status)
-    if [ "$STATUS" == "on" ]; then
-        VAL=$(mixer_info | sed -n -E "s/.*\[(-?[0-9]+.?[0-9]+)$UNIT\].*/\1/p" | head -n 1)
-        echo "$VAL$UNIT"
-    else
-        echo ""
-    fi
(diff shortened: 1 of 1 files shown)


kgilmer:
Looks good @farsil , and you briefly mention what you did to test your changes?

farsil:
I am currently using in my machine with no problems. Later this afternoon I can try installing it into the virtual machines (Ubuntu 20.04 + 1.4 and Regolith 1.4 Xubuntu-based) to check if there's something missing.

kgilmer:
If your machine is running 1.4 that should be sufficient, thanks :)

kgilmer:
I'm about to make a release of this package and will take this as part of it.

PR #49, opened 2020-05-07
Update keyboard-layout

kirushik:
GSettings has a special option, `mru-sources` (as in Most Recently Used sources) which is specified to change order with every layout change.
The original version uses `sources` which doesn't guarantee any reordering, and didn't work reliably for me (on Ubuntu 20.04)

**PS.** Probably adding an explicit license to this repository would be a good idea; just in case, I explicitly grant all rights to this contribution of mine to Regolith project.

### DIFF
--- a/scripts/keyboard-layout
+++ b/scripts/keyboard-layout
@@ -6,5 +6,5 @@ VALUE_FONT=${font:-$(xrescat i3xrocks.value.font)}
 PANGO_START="<span color=\"${VALUE_COLOR}\" font_desc=\"${VALUE_FONT}\">"
 PANGO_END="</span>"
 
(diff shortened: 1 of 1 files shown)


kgilmer:
Thank you @kirushik !  I have added a license file that points to the debian copyright file for details.  Please let me know if you see something wrong w/ that approach.

kirushik:
> Please let me know if you see something wrong w/ that approach.

It might work, if that's what you prefer.
Otherwise, I've seen a lot of project putting the verbatim GPL3 text into `LICENSE` file in the repo, and tracking contributions in a separate `CONTRIBUTORS` file.

I can't tell you if that's better or worse than what you're doing now — on one hand, having those two "sources of truth" desynchronized might really be a hard to fix issue; on another, there's a definite benefit of having Github explicitly show your license in its UI and API, for contributors' peace of mind if not for anything else.

Also — thank you for your work on Regolith!

PR #53, opened 2020-05-15
Add option for net-traffic to display data in bits rather than bytes.

cjohnston1158:
Adapted from:

https://github.com/marssantoso/nettraffic/commit/c2377279ec8f0d4c287c2dd2777c8d720e6ca2c2


kgilmer:
Thank you @cjohnston1158 !

PR #55, opened 2020-05-22
Fixed missing disk-capacity indicator

moritzheiber:
- Resolved numerous lintian errors
- Added .gitattributes for easier packaging
- Fixed disk-capacity label not showing up
- Fixed missing color separation for disk-capacity output
- Bumped debhelper compatibility to 12
- Fixed missing or superfluous copyright information
- Removed whitespaces from debian/{changelog,rules} (lintian warning)

I'll leave the bumping of the `changelog` in your capable hands :wink:


moritzheiber:
@kgilmer You might want to make it a habit for yourself to run [lintian](https://lintian.debian.org/) on the Regolith packages regularly :smirk:

kgilmer:
Wow, this is a lot of cleanup work, thank you @moritzheiber .  A lot of this is remaining noise from the i3blocks fork.

Other than the `lintian` that's run as part of `debuild`, how do you suggest I use it?

.gitattributes
```diff
@@ -0,0 +1 @@
+debian export-ignore
```
kgilmer, line 1: I see the documentation for this but it's not clear to me what impact that this has?

moritzheiber: I used the following command to collect the `orig` tarball for testing the packaging:
```
$ git archive --format=tar.gz -o ../regolith-i3xrocks-config_$(dpkg-parsechangelog --show-field Version | cut -d'-' -f1).orig.tar.gz HEAD
```
it makes sure you only add files to the tarball that have been checked into git and discards any items you're "ignoring" via the attributes :slightly_smiling_face: 

I can remove it if you want to, it doesn't hurt though because it's only relevant to `git archive`.

moritzheiber:
> Wow, this is a lot of cleanup work, thank you @moritzheiber . A lot of this is remaining noise from the i3blocks fork.
> 
> Other than the `lintian` that's run as part of `debuild`, how do you suggest I use it?

I think it's mainly about adding `--pedantic` to the `lintian` run, which fails on errors. And, obviously, resolving at least the warnings (if not all the items) that pop up.

.gitattributes
kgilmer: Ah,  I like it!   I've been brutishly doing stuff like `tar cfzv ../<pkg filename> --exclude debian --exclude .git\* --exclude .idea --exclude .vscode .`

Thanks!

PR #56, opened 2020-05-23
Fix button clicks blocklets

moritzheiber:
- Also fix network bandwidth calculation/shellcheck errors
  in net-traffic script

Fixes https://github.com/regolith-linux/regolith-desktop/issues/336

### DIFF
--- a/scripts/net-traffic
+++ b/scripts/net-traffic
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -Eeu -o pipefail
+
 # span for text
 fspan() {
     echo "<span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\"> ${1}</span>"
@@ -21,7 +23,21 @@ GET="cat /proc/net/dev"
 # some default values for arguments"
 UP=${label_icon:-$(xrescat i3xrocks.label.up " up")}
 DN=${label_icon:-$(xrescat i3xrocks.label.dn " dn")}
+BUTTON=${button:-}
+DLY=${DLY:-1}
+RT=${RT:-}
+BPS=false
+UNIT=KB
 ULEV=2
+
+# Handle button event
+if [ "x${BUTTON}" == "x1" ] ; then
+    if [ "$INT_TYPE" == "W" ]; then
+        /usr/bin/i3-msg -q exec /usr/bin/gnome-control-center wifi
+    else
+        /usr/bin/i3-msg -q exec /usr/bin/gnome-control-center network
+    fi
+fi
 
 # determine the net interface name
 IF=$(ip route get 1.1.1.1 | grep -Po '(?<=dev\s)\w+' | cut -f1 -d ' ')
@@ -97,7 +113,7 @@ if [ "$UNIT" = "MB" ] || [ "$UNIT" = "M" ]; then
         4) SUF="Mb/s";;
         *) SUF="MB/s";;
     esac
-    elif [ "$UNIT" = "KB" ] || [ "$UNIT" = "K" ]; then
+elif [ "$UNIT" = "KB" ] || [ "$UNIT" = "K" ]; then
     FAC=1024
     case "$ULEV"
         in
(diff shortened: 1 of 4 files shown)


cheginit:
We're planning to add a floating window class to the i3 config file called `floating_window`. So for the button section, you can add `class=floating_window` flag to `gnome-control-center` to make the window floating.

See this [PR](https://github.com/regolith-linux/regolith-i3-gaps-config/pull/16/commits/bdf71bc31c132de029fcca71dbe5c2caf4406066).

moritzheiber:
@cheginit For me the window is floating when run from the blocklet, though, likely because [there's an option which applies to all GNOME Control Center windows](https://github.com/regolith-linux/regolith-i3-gaps-config/blob/master/config#L332).. is setting the explicit class even necessary then?

cheginit:
We're planning to replace it with this approach to make it more general so users can make any window floating by using this flag. It's a gtk flag so all gtk apps that implement it can be made floating.

moritzheiber:
> We're planning to replace it with this approach to make it more general so users can make any window floating by using this flag. It's a gtk flag so all gtk apps that implement it can be made floating.

Added the window class to all calls to `gnome-control-center`

scripts/net-traffic
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -Eeu -o pipefail
```
kgilmer, line 3: Interesting!  What's going on here?

scripts/net-traffic
```diff
@@ -21,8 +23,22 @@ GET="cat /proc/net/dev"
...
+BUTTON=${button:-}
+DLY=${DLY:-1}
+RT=${RT:-}
+BPS=false
+UNIT=KB
 ULEV=2
 
+# Handle button event
+if [ "x${BUTTON}" == "x1" ] ; then
+    if [ "$INT_TYPE" == "W" ]; then
```
kgilmer, line 35: Nice!

scripts/net-traffic
```diff
@@ -143,44 +154,6 @@ RX=$(echo "scale = 0; ($RX2-$RX1)/$DLY" | bc -l)
 TX=$(echo "scale = 0; ($TX2-$TX1)/$DLY" | bc -l)
 AX=$(echo "scale = 0; $RX+$TX" | bc -l)
 
-# set net unit automatically
```
kgilmer, line 146: This can be removed because it's redundant?

scripts/wifi
```diff
@@ -1,16 +1,17 @@
...
 
+BUTTON=${button:-}
 VALUE_COLOR=${color:-$(xrescat i3xrocks.value.color "#D8DEE9")}
 LABEL_ICON=${label_icon:-$(xrescat i3xrocks.label.wifi )}
 LABEL_COLOR=${label_color:-$(xrescat i3xrocks.label.color "#7B8394")}
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font "Source Code Pro Medium 13")}
 
 ESSID=$(LC_ALL=C nmcli -t -f active,ssid dev wifi |  perl -n -e '/yes:(.*)/ && print $1')
-if [ ! -z "$ESSID" ]; then
+if [ -n "$ESSID" ]; then
```
kgilmer, line 10: I'm learning!

scripts/net-traffic
moritzheiber: Yes, I’m getting rid of it in favor of a sane default (which also fixes a shellcheck error)

scripts/net-traffic
moritzheiber: `-E` makes sure we're inheriting errors from subshells
`-e` exits the script immediately upon an error (i.e. it doesn't "bleed" its errors either into userspace or, worse, the bar)
`-u` treats unset variables as errors (i.e. you have to initialize them or the script won't run)
`-o pipefail` fails the script with the status of the last pipeline execution (and not of the overall script)

moritzheiber: Generally, these are safe to use when you're running a properly written script

scripts/net-traffic
moritzheiber: We could keep both around though, if you prefer to have the code instead of the (arguable, forced) default

jefft:
Ahem, commit  8e7daa1 left the 'x' flag on in the net-traffic script, and now my `i3-gnome-flashback.desktop` systemd log is being filled with debug logs :)

kgilmer:
Thanks for helping us test in `unstable` @jefft :smile: .  We'll get that fixed up shortly.

PR #58, opened 2020-05-23
Added bluetooth blocklet

### LINKED ISSUE #57: Add a bluetooth blocklet
Add a blocklet for connected Bluetooth devices.

akss13:
python2 dependencies: [**bluetool**](https://pypi.org/project/bluetool/)
Fixes #57

### DIFF
--- a/scripts/bluetooth
+++ b/scripts/bluetooth
@@ -9,15 +9,12 @@ no_el=int(len(bluetooth.get_connected_devices()))
 fs_len=len(str(bluetooth.get_connected_devices()))
 
 place=fs_s.find('name') 
-#print(place)
-#print(fs_s[place+8])
 
 res = [i for i in range(len(fs_s)) if fs_s.startswith("name", i)]
 #print(res)
 l_r=len(res)
 d_name=""
 for x in range(0, l_r):
-	#print(fs_s[res[x]+8])
 	start=res[x]+8
 	end=res[x]+8
 	for y in range(start,fs_len):
@@ -26,13 +23,11 @@ for x in range(0, l_r):
 			break
 	for z in range(start,end):
 		d_name+=fs_s[z]
-	#d_name+=" "
+	d_name+=" "
 
 
 cmd='bluetooth | grep -wo \'on\' | tr \'\\n\' \'\\0\' || true'
 result = subprocess.check_output(cmd, shell=True)
-#print(result)
-#print(len(d_name))
 len_dname=len(d_name)
 if result=='on\0':
 	if len_dname!=0:
@@ -40,5 +35,3 @@ if result=='on\0':
 
 cmd_button="if [ ! -z \"$button\" ]; then /usr/bin/i3-msg -q exec /usr/bin/gnome-control-center bluetooth; fi"
 result = subprocess.check_output(cmd_button, shell=True)
-
-
--- a/scripts/bluetooth
+++ b/scripts/bluetooth
@@ -8,71 +8,6 @@ import sys
 import subprocess
 import os
 
-SERVICE_NAME = "org.bluez"
-ADAPTER_INTERFACE = SERVICE_NAME + ".Adapter1"
-DEVICE_INTERFACE = SERVICE_NAME + ".Device1"
-
-
-class BluezUtilError(Exception):
-    pass
-
-
-def get_managed_objects():
-    bus = dbus.SystemBus()
-    manager = dbus.Interface(
-        bus.get_object("org.bluez", "/"),
-        "org.freedesktop.DBus.ObjectManager")
-    return manager.GetManagedObjects()
-
-
-def find_adapter(pattern=None):
-    return find_adapter_in_objects(get_managed_objects(), pattern)
-
-
-def find_adapter_in_objects(objects, pattern=None):
-    bus = dbus.SystemBus()
-
-    for path, ifaces in objects.items():
-        adapter = ifaces.get(ADAPTER_INTERFACE)
-
-        if adapter is None:
-            continue
-
-        if not pattern or pattern == adapter["Address"] or path.endswith(
-                pattern):
-            obj = bus.get_object(SERVICE_NAME, path)
-            return dbus.Interface(obj, ADAPTER_INTERFACE)
-
-    raise BluezUtilError("Bluetooth adapter not found")
-
-
-def find_device(device_address, adapter_pattern=None):
-    return find_device_in_objects(
-        get_managed_objects(), device_address, adapter_pattern)
-
-
-def find_device_in_objects(objects, device_address, adapter_pattern=None):
-    bus = dbus.SystemBus()
-    path_prefix = ""
-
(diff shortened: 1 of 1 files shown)


scripts/bluetooth
```diff
@@ -0,0 +1,44 @@
...
+cmd='bluetooth | grep -wo \'on\' | tr \'\\n\' \'\\0\' || true'
+result = subprocess.check_output(cmd, shell=True)
+#print(result)
+#print(len(d_name))
+len_dname=len(d_name)
+if result=='on\0':
+	if len_dname!=0:
+		print("  "+d_name)
+
+cmd_button="if [ ! -z \"$button\" ]; then /usr/bin/i3-msg -q exec /usr/bin/gnome-control-center bluetooth; fi"
```
moritzheiber, line 41: Can we maybe do this natively in Python instead of shelling out? Also, please make sure the blocklet is only reacting to a left-click and nothing else. We have an outstanding PR for adjusting this for other blocklets too :slightly_smiling_face:

scripts/bluetooth
```diff
@@ -0,0 +1,44 @@
...
+import subprocess
+from bluetool import Bluetooth
+bluetooth = Bluetooth()
+
+fs_s=str(bluetooth.get_connected_devices())
+no_el=int(len(bluetooth.get_connected_devices()))
+fs_len=len(str(bluetooth.get_connected_devices()))
+
+place=fs_s.find('name') 
+#print(place)
```
moritzheiber, line 12: Are these comments required?

scripts/bluetooth
```diff
@@ -0,0 +1,44 @@
...
+	for y in range(start,fs_len):
+		if fs_s[y]=="\'":
+			end=y
+			break
+	for z in range(start,end):
+		d_name+=fs_s[z]
+	#d_name+=" "
+
+
+cmd='bluetooth | grep -wo \'on\' | tr \'\\n\' \'\\0\' || true'
```
moritzheiber, line 32: If we could stick with "native" Python here I'd prefer it ..

moritzheiber:
Additionally, the Python module you're using hasn't seen an update since 3 years and is only tested with Python 2.x .. which is deprecated. Maybe you want to use https://github.com/pybluez/pybluez instead?

akss13:
> Could you submit a small screenshot of how the blocklet looks? 
> 
> Another question, in general: What happens if somebody has several Bluetooth devices connected? How will it be displayed on the blocklet?

Screenshot from 2020-05-23 23-58-46
Screenshot from 2020-05-23 23-52-39

scripts/bluetooth
akss13: I actually really like the ability to open up the settings by scrolling. I'll try to make it work so that it works by left-clicking only.
I could not found a way to emulate the button in python natively.

scripts/bluetooth
akss13: Cleaned up the script.

scripts/bluetooth
akss13: I'll try to make the required changes for this shell command.

akss13:
> Additionally, the Python module you're using hasn't seen an update since 3 years and is only tested with Python 2.x .. which is deprecated. Maybe you want to use https://github.com/pybluez/pybluez instead?

I actually did know this existed. I'll replace the current one with and update the script.

scripts/bluetooth
moritzheiber: You basically just have to read the environment variable and then send a command to i3 using IPC. We're using [i3pc](https://pypi.org/project/i3ipc/) in [another blocklet](https://github.com/regolith-linux/regolith-i3xrocks-config/blob/d6555032e2de3889eb0eff89b621b4946e807694/scripts/focused-window-name) for this already

akss13: Thanks for the reference. I'll look into the example and implement it in the script.

akss13:
> Additionally, the Python module you're using hasn't seen an update since 3 years and is only tested with Python 2.x .. which is deprecated. Maybe you want to use https://github.com/pybluez/pybluez instead?
The module [pybluez](https://github.com/pybluez/pybluez) seems to have no functions for retrieving connected devices as per their [docs](https://pybluez.readthedocs.io/en/latest/api/index.html). So it would not be possible to change the module from [bluetool](https://pypi.org/project/bluetool/)

moritzheiber:
Bummer. Sorry, I didn't really read the API documentation before suggesting it. My fault entirely.

This is the code that's reading the device state: https://github.com/emlid/bluetool/blob/master/bluetool/bluetool.py#L89-L142

Maybe you can port it to Python3, directly into the blocklet? The only dependency then would be `python3-dbus` :slightly_smiling_face:

akss13:
> Bummer. Sorry, I didn't really read the API documentation before suggesting it. My fault entirely.
> 
> This is the code that's reading the device state: https://github.com/emlid/bluetool/blob/master/bluetool/bluetool.py#L89-L142
> 
> Maybe you can port it to Python3, directly into the blocklet? The only dependency then would be `python3-dbus`

Converted the script to python3 :smile: . 
Also made changes so that the settings open by left-clicking only.

moritzheiber:
@akss13 Great effort, thank you. Right now you're covering all the use-cases, i.e. also `Available` and `Paired` instead of "just" `Connected`. That means there's also _a lot_ of code in there that's not needed. Can we strip it away to only retain the code we need in order to get all connected devices?

akss13:
> @akss13 Great effort, thank you. Right now you're covering all the use-cases, i.e. also `Available` and `Paired` instead of "just" `Connected`. That means there's also _a lot_ of code in there that's not needed. Can we strip it away to only retain the code we need in order to get all connected devices?

Trimmed unnecessary parts from the script.

moritzheiber:
I also think we don't need these lines: https://github.com/regolith-linux/regolith-i3xrocks-config/pull/58/files#diff-1ce3b03cd4e9d668f6335bd6efeaf825R46-R59

akss13:
> I also think we don't need these lines: https://github.com/regolith-linux/regolith-i3xrocks-config/pull/58/files#diff-1ce3b03cd4e9d668f6335bd6efeaf825R46-R59

Sorry overlooked those lines :sweat_smile: . I guess it's ok now

moritzheiber:
Fantastic! Now, the only thing left is to run [`flake8`](https://pypi.org/project/flake8/) on it once, fix the linting issues and then we're ready to merge!

akss13:
I've run linter and flake8 on it.
Just last-minute thoughts, I'm using Font-awesome glyph i.e, ``. It may cause issues since by default Material Design Icons font come installed.

moritzheiber:
@akss13 Great effort, thank you. Should there be any issues in the future we'll resolve them then :slightly_smiling_face:

kgilmer:
Yes, thank you for the contribution @akss13 !

kgilmer:
This change is building into `unstable` now.

PR #63, opened 2020-05-26
Add option to define a specific media player

joelhoisko:
This way you can define your own specific media player to show information for with the Xresource `i3xrocks.media-player`.

My reason for this was I didn't want my youtube tabs be considered media players (they are listed in `playerctl -l` output) and I only wanted the tray to show the output for my spotify. Using `playrctl` without the `--player=` option defaults to the first in the list, which was something like `chrome.instance56851` for me usually.

Other way which we could do this would be to allow the user to add a property for the media-player block, like in the [i3blocks-contrib](https://github.com/vivien/i3blocks-contrib/tree/master/mediaplayer) version.

I also didn't know how to actually build this locally so I haven't tested if it actually works, the string produced by the commands should be correct.

Inspired by [mediaplayer](https://github.com/vivien/i3blocks-contrib/blob/d600a2c481e489bacf1118e68063ccf3750da4a1/mediaplayer/mediaplayer#L32) from i3blocks-contrib.


moritzheiber:
I like the idea, but maybe we should just make it possible to pass arguments? Without a specific `MEDIA_PLAYER` variable? It would still fit your request and it would be a little bit more transparent towards the user

kgilmer:
In that case @moritzheiber the media player would be defined in the user copy of `/etc/regolith/i3xrocks/conf.d/xx_media-player` @moritzheiber ?

moritzheiber:
@kgilmer No, I meant, just let users straight out define the `PLAYER_ARGS` variable via a Xresource (e.g. `i3xrocks.player_args`), which would make it easier for folks to add their own options _and_ satisfy the author's requirements :slightly_smiling_face:

kgilmer:
Oh, right got it.  Less is more.  :+1:

joelhoisko:
I made it so to read the args straight from `i3xrocks.media_player.player_args`. I like the idea and how it gives users more control over the script but not all `playerctl` options are really applicable here so some documentation on how this feature works would be nice too. Is there some sort of documentation page/readme to describe all these scripts and their (possible) xresources?

kgilmer:
Good point @joelhoisko...there is no documentation per-blocklet other than comments in the config stanzas ([example](https://github.com/regolith-linux/regolith-i3xrocks-config/blob/master/conf.d/30_net-traffic)) but we do have a reference page for all Xresource values.  If you add your keys as a PR [here](https://github.com/regolith-linux/website/blob/master/content/en/docs/Reference/xresources.md), it will get published with your enhancement.  :smile:

scripts/media-player
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
+# for storing possible user defined arguments for playerctl
+PLAYER_ARGS=$(xrescat i3xrocks.media_player.player_args)
```
kgilmer, line 3: I want to verify, this will also work if `i3xrocks.media_player.player_args` is undefined or should it default to `status`.  If so you can just add `status` like so:
```
$(xrescat i3xrocks.media_player.player_args status)
```

kgilmer:
Any update on the status of this PR @joelhoisko ?

scripts/media-player
joelhoisko: I tested it out and it still works as usual aka taking the first player in the `playerctl -l` list if there's no Xresource defined. If the user puts some random mumbo jumbo to it then the script just dies but that's on the user.

The actual `status` command is executed on the the next line nevertheless of the given player args:
```bash
STATUS=$(playerctl $PLAYER_ARGS status 2> /dev/null)
```

joelhoisko:
Sorry for the long delay, summer and all. 

- Merged newest master in
- Added the `$PLAYER_ARGS` also into the default mouse click options (from https://github.com/regolith-linux/regolith-desktop/issues/352)
- Changed Xresource key naming to use normal dashes as that seems to be the norm eg. `i3xrocks.media-player.player-args`
- PR for docs https://github.com/regolith-linux/website/pull/40

kgilmer:
No worries regarding the delay and thanks for following up!  PR looks good!

joelhoisko:
Thanks for the awesome DE you're making ❤️

PR #64, opened 2020-05-26
reworked keyboard-layout script to reliably determine keyboard layout…

### LINKED ISSUE #52: keyboard layout issues
Hi, I noticed that the keyboard layout does not display properly in the bar if you only have one language setup on your computer. Looks like it is related to an issue with gnome where mru-sources is not initialized on new installs and is only used when multiple languages are installed on the system. I have a fix for this and can issue a pull request if you would like me to do so.

buck10:
… from gsettings. added handling of mouse clicks to open region settings and display keyboard layout

Fixes #52

### DIFF
--- a/conf.d/20_keyboard-layout
+++ b/conf.d/20_keyboard-layout
@@ -2,4 +2,4 @@
 #
 # Displays current keyboard layout.
 [keyboard-layout]
-interval=persist
+interval=1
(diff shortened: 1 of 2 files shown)


moritzheiber:
Thanks!

kgilmer:
Yes, well done.  thank you @buck10 !

PR #65, opened 2020-05-27
Default icon and fix dependecy

cheginit:


debian/control
```diff
@@ -172,7 +172,7 @@ Package: i3xrocks-todo
 Architecture: any
 Depends: ${misc:Depends},
     regolith-i3xrocks-config,
-    td-cli
+    python3-td-cli
```
kgilmer, line 175: I think this will change again.  Let's see if td-cli 1.2.4-7 works for you.

cheginit: Yes, I just removed `python3-td-cli` and installed `td-cli` and it works.

PR #68, opened 2020-06-24
Exit gracefully if the focused window does not have a name

cqql:
Some windows do not have names and every time I get a Ubuntu crash dialog for this script. Now the script checks if `name` is actually a string.


moritzheiber:
Much appreciated, thank you!

PR #71, opened 2020-08-17
Add Pango formatted special characters

leosunmo:
This will make sure the i3 pango formatter displays these special (in the Pango language) characters properly.

I only tested this with the ampersand (&) as I couldn't find any songs on Spotify with "<>" characters :shrug: 

This is using the same method I used in https://github.com/leosunmo/spotify-blocklet/blob/master/main.go#L114 which has worked great so far.

### DIFF
--- a/scripts/media-player
+++ b/scripts/media-player
@@ -37,6 +37,11 @@ else
     TEXT=$(printf "%-${width}s" "$text")
 fi
 
+# sanitize text for Pango parsing
+TEXT=${TEXT//&/&amp;}
+TEXT=${TEXT//>/&gt;}
+TEXT=${TEXT//</&lt;}
+
 echo "<span color=\"${LABEL_COLOR}\">${ICON}</span>\
 <span font_desc=\"${VALUE_FONT}\" color=\"${VALUE_COLOR}\"> \
 ${TEXT}</span>"


leosunmo:
There's probably other special characters that should be Pango-escaped actually. Might want to find an exhaustive list...

leosunmo:
A good one to test general character madness is of course good old trusty `⣎⡇ꉺლ༽இ•̛)ྀ◞ ༎ຶ ༽ৣৢ؞ৢ؞ؖ ꉺლ - ࿃ूੂ࿃ूੂੂ࿃ूੂOOOOOOOOOOOO ̟̞̝̜̙ ooooooooo`
https://open.spotify.com/artist/1TIbqr0x8HoKzKBNtNN8wf?si=mlCgetgrSVCwJWkDhqyu_g

leosunmo:
What is the process for the Changelog? Do I manually add an entry?

cheginit:
I think @kgilmer will add it to the deb package log file.

cheginit:
Also, one user on Slack suggested changing `if [[ $length > $width ]]; ` to `if (( $length > $width ));` or `if [[ $length -gt $width ]]; ` to support cutting off for more than 100 characters.

kgilmer:
changelog: yes that gets updated when I run `gbp dch` which takes the git changelog and puts it in the debian changelog.  And then I sign the source package and upload it to launchpad.  

LMK when this is ready merge.

leosunmo:
I'll add the length check in this PR as well so I don't have to open a new PR

leosunmo:
Okay so I went ahead and made more changes... if you want I can split this in to a new PR as it no longer fits within the pango fix. I might rename the PR name.

I added separate length parsing for artist and song. I prefer to cut both off independently, but that might just be me. This is how my own bar handles it. 

The current max width of 25 is a bit too short for this however, might want to look at increasing the default? Please give this a try and see if it's something that makes sense for you!

PR #76, opened 2020-10-06
Make curl work for special characters

pmahnke:
## Done

- Put curl for weather in double quotes as the `&format` wasn't being passed properly and returned multiple lines

## QA

- download this branch
- run `weather` from the command line
- see that the output is on one line


moritzheiber:
Great stuff, thanks!

Issue #77, opened 2020-10-10
Support for multiple batteries


pedrokost:
My laptop has two batteries, BAT0 and BAT1 (Lenovo T440s). The battery recognition system of `i3xrocks/battery` only takes the first battery (in my case the smaller one), which stays charged until the second, larger battery discharges.This makes the battery indicator very misleading.

I believe it would be better to do a weighted average of the battery percentages. Something like (BAT0.energy-full * BAT0.percentage + BAT1.energy-full * BAT1.percentage) / (BAT0.energy-full + BAT1.energy-full). Alternatitely, showing two battery icons would also be fine.

-----------

Some info:

```
upower -e
/org/freedesktop/UPower/devices/line_power_AC
/org/freedesktop/UPower/devices/battery_BAT0
/org/freedesktop/UPower/devices/battery_BAT1
/org/freedesktop/UPower/devices/DisplayDevice
```

```
❯ upower --show-info "/org/freedesktop/UPower/devices/battery_BAT0"
  native-path:          BAT0
  vendor:               LGC
  model:                45N1113
  serial:               8426
  power supply:         yes
  updated:              sob 10 okt 2020 23:50:56 (38 seconds ago)
  has history:          yes
  has statistics:       yes
  battery
    present:             yes
    rechargeable:        yes
    state:               discharging
    warning-level:       none
    energy:              14,94 Wh
    energy-empty:        0 Wh
    energy-full:         21,4 Wh
    energy-full-design:  23,48 Wh
    energy-rate:         10,904 W
    voltage:             11,527 V
    time to empty:       1,4 hours
    percentage:          69%
    capacity:            91,1414%
    technology:          lithium-ion
    icon-name:          'battery-full-symbolic'
  History (charge):
    1602366656	69,000	discharging
  History (rate):
    1602366656	10,904	discharging

❯ upower --show-info "/org/freedesktop/UPower/devices/battery_BAT1"
  native-path:          BAT1
  vendor:               LGC
  model:                45N1738
  serial:               203
  power supply:         yes
  updated:              sob 10 okt 2020 23:51:31 (8 seconds ago)
  has history:          yes
  has statistics:       yes
  battery
    present:             yes
    rechargeable:        yes
    state:               discharging
    warning-level:       low
    energy:              3,48 Wh
    energy-empty:        0 Wh
    energy-full:         63,54 Wh
    energy-full-design:  71,1 Wh
    energy-rate:         13,0444 W
[...]

PR #78, opened 2020-10-11
[IMPR] Show aggregated battery percentage

pedrokost:
This is a fix for issue #77 . It adds support for showing aggregate battery percentage for 1 or more batteries. It relies on upower's [DisplayDevice](https://upower.freedesktop.org/docs/UPower.html#id-1.2.3.8.5).

If a user prefers to show a specific battery (e.g. battery_BAT0) they can do so by setting `BLOCK_INSTANCE` to `battery_BAT0`.

The `get_battery_percent` function has also been improved to handle cases where the returned percentage is float with a comma used as a floating point separator.

### DIFF
--- a/scripts/battery
+++ b/scripts/battery
@@ -19,11 +19,12 @@ get_battery_status() {
 # Returns:
 #   The battery's percentage, without the %.
 get_battery_percent() {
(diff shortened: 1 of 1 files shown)


kgilmer:
Thank you @pedrokost :)

Issue #73, opened 2020-09-15
i3xrocks net-traffic appears and dissaperas on Ubuntu 20.04


nachovizzo:
I just upgraded to Ubuntu 20.04 and this is happening: [video](https://streamable.com/a3pq1y)

The only related issue I've found is on [reddit](https://www.reddit.com/r/i3wm/comments/fwj7r9/net_traffic_block_keeps_appearing_and_disappearing/)

Any idea on how to debug/solve this issue?

kgilmer:
I like the music!

As mentioned in the reddit post, run the net-traffic script from a terminal when it's not visible in the bar.  Generally speaking, what you see in the bar is simply output from the shell scripts in `/usr/share/i3xrocks`

You should either see the correct output, something like: 

```bash
$ /usr/share/i3xrocks/net-traffic 
<span color="#E6E1CF"></span><span font_desc="Source Code Pro Medium 13" color="#E6E1CF">  3.1KB</span><span color="#E6E1CF"></span><span font_desc="Source Code Pro Medium 13" color="#E6E1CF">  2.9KB</span><span color="#E6E1CF"></span>
```

Or something else that may just be error stuff.  This is what we need to determine why you're seeing this problem.

nachovizzo:
Thanks! the music is from Sig Ragga: https://open.spotify.com/artist/7hBKLnKOnLXNBwJ1l6p24J?si=-0iWGoheRaWIdtxfSQQ88Q.

I ran the `net-traffic` from the shell and when there is nothing in the bar, then, there is also nothing in the shell...

```sh
ivizzo (master) /usr/share/i3xrocks/net-traffic 
<span color="#e6e1cf"></span><span font_desc="Source Code Pro Medium 13" color="#e6e1cf">  6,0KB</span><span color="#e6e1cf"></span><span font_desc="Source Code Pro Medium 13" color="#e6e1cf">  5,0KB</span><span color="#e6e1cf"></span>
ivizzo (master) /usr/share/i3xrocks/net-traffic
ivizzo (master) /usr/share/i3xrocks/net-traffic
ivizzo (master) /usr/share/i3xrocks/net-traffic
ivizzo (master) /usr/share/i3xrocks/net-traffic
ivizzo (master) /usr/share/i3xrocks/net-traffic
[...]

PR #79, opened 2020-10-11
[FIX] Fixes dissapearing indicator when value is 0

pedrokost:
Fixes issue #73 by setting the value of RN, TN and AN to 0 instead of exiting if the value is 0 or invalid.


kgilmer:
Awesome, thanks for this PR @pedrokost .  I'm testing it locally now.

kgilmer:
No problems found in testing.  @pedrokost thank you for the fix!

PR #81, opened 2020-10-21
Fix crash when focused window name is None

lacerda:
Fixes the following crash:
```
Traceback (more recent call last):
  File "/usr/share/i3xrocks/focused-window-name", line 78, in <module>
    if "LibreOffice" in name.split("-")[-1] and instance == "libreoffice":
AttributeError: 'NoneType' object has no attribute 'split'
```


moritzheiber:
Thanks!

PR #82, opened 2020-10-23
reading from xresources now

ferdinandyb:
This pull request fixes the bug reported in https://github.com/regolith-linux/regolith-desktop/issues/470 as it will now use `xrdb -query` to read the workspace names directly from the configurations set by regolith. This also means we can deprecate the `--startnum` flag, since we can now read the smallest workspace number from the configs.

### DIFF
--- a/scripts/next-workspace
+++ b/scripts/next-workspace
@@ -2,28 +2,47 @@
 
 import i3ipc
 import argparse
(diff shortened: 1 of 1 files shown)


kgilmer:
Sorry, somehow I missed this PR.

PR #85, opened 2020-11-12
Restore original behavior of mouse actions on volume

twvd:
This was broken in https://github.com/regolith-linux/regolith-i3xrocks-config/commit/2bfab21c15a9d730aaf9885bd777122f2ef72c16#diff-18e70f7d8538e1b83ef4439d10ed0ac98c433f6551932b2d172544b23b84b637

Also add overrides for every possible action (left, middle, right mouse buttons and scrolling actions)

### DIFF
--- a/scripts/volume
+++ b/scripts/volume
@@ -72,13 +72,17 @@ mixer_value() {
     fi
 }
 
-ACTION_1=$(xrescat i3xrocks.action.volume.left "amixer -q -D $MIXER sset $SCONTROL $(mixer_capability) toggle")
-ACTION_2=$(xrescat i3xrocks.action.volume.middle "amixer -q -D $MIXER sset $SCONTROL $(mixer_capability) ${STEP}${UNIT}+")
-ACTION_3=$(xrescat i3xrocks.action.volume.right "amixer -q -D $MIXER sset $SCONTROL $(mixer_capability) ${STEP}${UNIT}-")
+ACTION_1=$(xrescat i3xrocks.action.volume.left "true")
+ACTION_2=$(xrescat i3xrocks.action.volume.middle "true")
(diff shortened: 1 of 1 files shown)


moritzheiber:
Thank you! Could you open a PR for adding the relevant documentation as well?

scripts/volume
```diff
@@ -72,13 +72,17 @@ mixer_value() {
     fi
 }
 
-ACTION_1=$(xrescat i3xrocks.action.volume.left "amixer -q -D $MIXER sset $SCONTROL $(mixer_capability) toggle")
-ACTION_2=$(xrescat i3xrocks.action.volume.middle "amixer -q -D $MIXER sset $SCONTROL $(mixer_capability) ${STEP}${UNIT}+")
-ACTION_3=$(xrescat i3xrocks.action.volume.right "amixer -q -D $MIXER sset $SCONTROL $(mixer_capability) ${STEP}${UNIT}-")
+ACTION_1=$(xrescat i3xrocks.action.volume.left "true")
```
kgilmer, line 75: I must be missing something but it's not clear to me how `ACTION_1` set to `true` causes any action to occur in the eval expression below.  The old code executes the `amixer` command.

kgilmer: Oh, I think the intention is to not perform any action by default for left or middle click.

kgilmer: Right, it's clear from the linked to diff.  Perfectly clear, thanks for your contribution @twvd!

twvd: Yeah, true is just ment to be a no-op. The original code had mute on right click and it moved in your change, so I moved it back to the original button.

twvd:
@moritzheiber Sure, I will submit a PR with documentation for the website one of these days.

PR #86, opened 2020-11-12
Fix for button handling of the net-traffic script

moritzheiber:
Previously, the button handling was run before the variable `INT_TYPE` was initialized. This should've been detected earlier (especially by `shellcheck`) but wasn't. With the recently added tests the script should now run "flawlessly".

I've also exchanged the default route parsing logic for a shorter, more efficient one-liner.

### DIFF
--- a/scripts/net-traffic
+++ b/scripts/net-traffic
@@ -30,19 +30,8 @@ UNIT=${UNIT:-}
 BPS=false
 ULEV=2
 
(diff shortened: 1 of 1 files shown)


moritzheiber:
Woo, the test has passed! :tada:

kgilmer (approved):
Fantastic, thanks!

PR #87, opened 2020-11-12
Make time interval configurable with Xresource

winder:
I forgot to push this config earlier. Now that i3xrocks **config files** are xresource-aware the bar can be configured like this.


moritzheiber:
(We can ignore the failing test for now, it's related to #86)

kgilmer:
Thanks for the follow up @winder !

PR #89, opened 2020-11-29
format the connected device string using pango same as other rocks

roxchgt:
well, bluetooth text didn't obey the formatting of other bar elements!
now it will

### DIFF
--- a/scripts/bluetooth
+++ b/scripts/bluetooth
@@ -68,6 +68,11 @@ class Bluetooth(object):
 
 ########################################
 
+typeface = subprocess.check_output(
+                        "/usr/bin/xrescat i3xrocks.value.font",
+                        shell=True,
+                        universal_newlines=True
+                        )
 
 bluetooth = Bluetooth()
 fs_s = str(bluetooth.get_connected_devices())
@@ -91,7 +96,7 @@ for x in range(0, l_r):
 
 len_dname = len(d_name)
 if len_dname != 0:
-    print("  "+d_name)
+    print('<span font_desc="'+typeface+'">'+'  '+d_name+'</span>')
 
 env_button = os.environ.get('button')
 cmd_open = "/usr/bin/i3-msg -q exec /usr/bin/gnome-control-center bluetooth"
--- a/scripts/bluetooth
+++ b/scripts/bluetooth
@@ -97,12 +97,12 @@ for x in range(0, l_r):
         d_name += fs_s[z]
     d_name += " "
 
-pango_formatted_output="<span color=\"{valueColor}\" font_desc=\"{typeface}\">\
-<span color=\"{labelColor}\">  </span>{deviceName}</span>"
+pango_output="<span color=\"{labelColor}\">  </span>\
+<span color=\"{valueColor}\" font_desc=\"{typeface}\">{deviceName}</span>"
 
 len_dname = len(d_name)
 if len_dname != 0:
-    print(pango_formatted_output.format(
+    print(pango_output.format(
     valueColor = valueColor,
     labelColor = labelColor,
     typeface = typeface,
--- a/scripts/bluetooth
+++ b/scripts/bluetooth
@@ -69,10 +69,13 @@ class Bluetooth(object):
 ########################################
 
 typeface = subprocess.check_output(
-                        "/usr/bin/xrescat i3xrocks.value.font",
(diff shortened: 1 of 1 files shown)


roxchgt:
@kgilmer Is it gonna be reviewed?

kgilmer:
Hi @roxchgt , thanks for submitting the PR.  You're definitely going in the right direction with this change. :smile:   To make the pango markup fully consistent with the other rocks, we need two `span`s, one for the icon and one for the text.  Like this:

```
<span color="#E6E1CF"></span><span font_desc="JetBrains Mono Medium 13" color="#E6E1CF"> 05/30 19:06 </span>
```

Please see [this page](https://regolith-linux.org/docs/developer/status-indicators/) for complete details.  Can you update the PR to work this change in?

roxchgt:
Done! I was focused on typeface, since the style I use don't have different colors for label and value. But now it's fully compatible, hopefully!
hopefully, because I ignored the fallback typeface and colors if xrescat doesn't return anything…! I assumed there's always a valid style set

scripts/bluetooth
```diff
@@ -94,9 +97,17 @@ for x in range(0, l_r):
         d_name += fs_s[z]
     d_name += " "
 
+pango_formatted_output="<span color=\"{valueColor}\" font_desc=\"{typeface}\">\
```
kgilmer, line 100: I believe the colors and fonts are in the wrong spot here.  For example when I run the time block: 

```bash
$ /usr/share/i3xrocks/time 
<span color="#E6E1CF"></span><span font_desc="JetBrains Mono Medium 13" color="#E6E1CF"> 12/07 10:16 </span>
```

Notice that the first span is for the color of the icon (the font is defaulted in the i3 config), and the second span specifies the non-icon font and the color of the text.

kgilmer (changes requested):

scripts/bluetooth
roxchgt: Oh I see… I though that I'd set the same font for both label and value and only override the color of label span,
like (in pseudo code) `<span font=font color=value.color><span color=label.color>label</span> label</span>`
but well, it could break the look if the user's selected font doesn't support labels.

kgilmer:
This is great, I've tested it locally and works very well.  There is one more thing to make it perfect, read the icon character also from Xresources rather than just using the hard-coded value in the script.  Here is an example: https://github.com/regolith-linux/regolith-i3xrocks-config/blob/master/scripts/todo#L7.  Your changes are an improvement over what was there before.  We can merge this PR as-is or you if you like you can define a new Xres key for the bluetooth status indicator and add the code to read that from Xres or default to the character in there if it's not defined in Xresources.  What would you like to do @roxchgt ?

roxchgt:
> This is great, I've tested it locally and works very well. There is one more thing to make it perfect, read the icon character also from Xresources rather than just using the hard-coded value in the script. Here is an example: https://github.com/regolith-linux/regolith-i3xrocks-config/blob/master/scripts/todo#L7. Your changes are an improvement over what was there before. We can merge this PR as-is or you if you like you can define a new Xres key for the bluetooth status indicator and add the code to read that from Xres or default to the character in there if it's not defined in Xresources. What would you like to do @roxchgt ?

so I've changed it so that it reads the icon from Xresources if it exists and if not, the default or fallback will be used. In order to add the label key I'll have to fork the other repo as well, am I right? How both changes in 2 repos are gonna be related?!

kgilmer:
> so I've changed it so that it reads the icon from Xresources if it exists and if not, the default or fallback will be used.
Great, thank you @roxchgt :smile: 

> In order to add the label key I'll have to fork the other repo as well, am I right? 

Well, that can certainly be done but it's not necessary.  Many blocks rely on hard-coded defaults, but by adding the check for Xresources, it allows users to easily specify their own icon choices if they prefer.  What you've done here is sufficient for that.

> How both changes in 2 repos are gonna be related?!

Yes they wouldn't really be related in any way except adding links to each of the PRs from the other.  But, as this is just overriding the default it's not a problem if one PR gets merged without the other.

kgilmer:
Your change is building into the `unstable` PPA now: https://launchpad.net/~regolith-linux/+archive/ubuntu/unstable/+sourcepub/11817522/+listing-archive-extra

PR #90, opened 2020-12-04
Match example with what the script is looking for

garyritchie:
The script is looking for uppercase MNT, not lowercase mnt.


moritzheiber:
Thanks!

PR #92, opened 2020-12-07
Rewrote net-traffic script, fixes #91

### LINKED ISSUE #91: i3xrocks-net-traffic shows 0kb/0kb when there is traffic
- [x] I've [run a search on the existing bug reports](https://help.github.com/en/github/searching-for-information-on-github/searching-issues-and-pull-requests) to make sure I'm not filing a duplicate (you can use the search box at the top left corner)
- [ ] I've read [the Regolith documentation](https://regolith-linux.org/docs/) and didn't find any relevant topic
- [x] I tried [asking on Slack](https://regolith-linux.org/interact/) (especially for configuration related issues)

In some situations i3xrocks-net-traffic shows 0kb/0kb when there is traffic. It is observed when on a Zoom session, and possibly also when streaming video and when sending data via UDP.

**To Reproduce**
Run a Zoom session or run `iperf3 -c ping.online.net -p 5203 -u` which will send via UDP. If removing `-u` it send via TCP, which does not seem to be affected.

**Expected behavior**
i3xrocks-net-traffic shows amount of traffic transmitted.

**Screenshots**
https://ibb.co/YDqjG2V

**Installation Details**
 - Regolith Install Type: ISO
 - Regolith Version: 1.5
 - PPA url: ppa:regolith-linux/stable
 - Host OS (for PPA): e.g. Ubuntu 20.10

moritzheiber:
Completely rewrote the net-traffic script (well, almost) to use `numfmt` instead of coming up with our own routine. Also uses `sysfs` instead of `proc` filtering now to determine the device type (with potential to add more devices) and traffic deltas.

Also added tests, which are regexp-based for now, but better than nothing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: |
           apt update -qq
-          apt install --no-install-recommends -y iproute2 bc
+          apt install --no-install-recommends -y iproute2 bc silversearcher-ag
           cp tests/xrescat_mock /usr/bin/xrescat
           cp tests/i3-msg_mock /usr/bin/i3-msg
       - name: Run net-traffic test
--- a/conf.d/30_net-traffic
+++ b/conf.d/30_net-traffic
@@ -1,5 +1,3 @@
 # Displays network load
 [net-traffic]
-# Uncomment the next line to display in bits
-# command=/usr/share/i3xrocks/net-traffic-new -b
 interval=10
--- a/scripts/net-traffic
+++ b/scripts/net-traffic
@@ -6,7 +6,7 @@ set -Eeu -o pipefail
 UP=${label_icon:-$(xrescat i3xrocks.label.up " up")}
 DN=${label_icon:-$(xrescat i3xrocks.label.dn " dn")}
 BUTTON=${button:-}
-DLY=${DLY:-1}
+DLY=${DLY:-5}
 RT=${RT:-}
 
 # determine the net interface name
--- a/scripts/net-traffic
+++ b/scripts/net-traffic
@@ -2,6 +2,55 @@
 
 set -Eeu -o pipefail
 
+# some default values for arguments"
+UP=${label_icon:-$(xrescat i3xrocks.label.up " up")}
+DN=${label_icon:-$(xrescat i3xrocks.label.dn " dn")}
+BUTTON=${button:-}
+DLY=${DLY:-1}
+RT=${RT:-}
+
+# determine the net interface name
+IF="${BLOCK_INSTANCE:-$(ip route show default | awk '{ print $5 }')}"
+[ "$IF" = "" ] && exit
+
+IF_PATH="/sys/class/net/${IF}"
+
+if [ -d "${IF_PATH}/wireless" ] ; then
+  NIC=${label_icon:-$(xrescat i3xrocks.label.wifi )}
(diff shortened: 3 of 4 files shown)


moritzheiber:
This should fix #91

moritzheiber:
There are some inconsistency in the way we're calling the script (`interval` is set to 10 seconds but the gathering period is only 1 second, meaning we're "losing" out on 9 seconds of traffic potentially), but that's only going see an improvement if we switch to `repeat` (which does work but produces a lot of noise), `persist` (which would mean we'd have to re-architect the script to run in an infinite loop) or bump the delay between measurements to something like `5`.

The last comment in #91 says the issue is fixed by this PR.

I would suggest bumping the delay to `5` and call it a day.

@kgilmer Opinions?

moritzheiber:
We could also switch to `repeat` and bump the delay to 5 seconds .. which would lower the noise but produce more accurate readings ..

kgilmer:
Awesome!  Quite a reduction in code/complexity, great work.  Regarding the delay, 5 seconds sounds fine to me.  Regarding the timing specifics, accuracy is good!  If there is some trade off between the various timing options you listed here, please LMK.  I'm going to pull your branch and do a basic sanity test.

kgilmer (approved):
Tested locally, works great.  LGTM

moritzheiber:
`repeat` vs. keeping the interval at `10` will increase the times the script is run, e.g. while it only runs for 2 seconds every 10 seconds it'll then run for 5-6 seconds straight, every 5-6 seconds (i.e. it'll "always" run). This could, potentially, have an impact on battery life, however, since we're not exactly doing rocket science here (especially not in bash) and we're sleeping for 5 seconds out of those 5-6 seconds, I'd say that's not really an issue.

@kgilmer If you agree I'll push the relevant changes to the branch and then merge them for packaging

kgilmer:
sounds good @moritzheiber !

moritzheiber:
@kgilmer Just discovered a major caveat (patting myself on the back for being diligent about reading documentation :sweat_smile:): https://github.com/vivien/i3blocks#interval

We're losing the button click functionality (which, right now, depending on interface type, launches GNOME Control Center's network tab) when we switch to `repeat` (which would be the best option for data accuracy). Are we okay with this?

Otherwise I'd retain the interval setting (`10`) and just bump the delay (to e.g. `5` or `7`), which does have the potential of introducing a race condition though (if `interval` is lower than `delay` the script will be spawned over and over again before returning a value)

cabunar:
If i have a multiple default route,  the script doesn't work.
For example, if i have a vpn connection:
`default via 10.0.0.1 dev tun0 proto static metric 50`
`default via 192.168.1.1 dev wlp2s0 proto dhcp metric 600`

`regolith.desktop[2140702]: cat: '/sys/class/net/tun0'$'\n''wlp2s0/statistics/rx_bytes': No such file or directory`

A possible solution:
`+IF="${BLOCK_INSTANCE:-$(ip route show default | awk '{ print $5 }' | head -n1)}"`

moritzheiber:
@cabunar I've addressed this by letting `awk` only read the first line. That should fix the issue you mentioned.

moritzheiber:
Making a judgement call here. Once the tests have passed I'll merge this with the delay set to 5 seconds instead of 1. We'll get more accurate readings and retain the button functionality. In the spirit of not breaking too much functionality I think that's a fair trade

kgilmer:
Interval has been changed to `repeat`.

PR #96, opened 2020-12-16
Rewrite weather blocklet

### LINKED ISSUE #95: Getting wrong weather forecast
Don’t know if this is the right place but the i3xrocks weather plugin gives me this URL for the weather forecast:

<https://wttr.in/Vienna,Vienna>

While it looks unsuspiciously on first sight, it actually is the Vienna in Clark County, South Dakota, 57271, United States of America – not the one I live in, the capital of Austria.

The string is composed via this line in  `/usr/share/i3xrocks/weather`:

```sh
IP_LOCATION="$(curl -s https://ipapi.co/city/),$(curl -s https://ipapi.co/region/)"
```

[Igor Chubin](https://twitter.com/p3k/status/1338485033351458816), the author of wttr.in gave some alternatives for naming a place, and so I tried it with `Austria,Austria,AT` which gives the [correct forecast](https://wttr.in/Vienna,Vienna,Austria)!

I.e. the line would work for me like this:

```sh
IP_LOCATION="$(curl -s https://ipapi.co/city/),$(curl -s https://ipapi.co/region/),$(curl -s https://ipapi.co/country)"
```

Now is the script handled by Regolith or would I have to turn to somewhere else to ask for this fix?

moritzheiber:
Here's a (opinionated) rewrite of the weather blocklet:

1. **Don't** do our own IP "magic", wttr.in does this for us, and it's much more reliable and way easier to do. If it doesn't detect it properly you can still set `weather_location` to something sensible.
2. **Don't** do our own temperature unit "magic", again, wttr.in will select the "right" unit for us, depending on the location you're using it from (e.g. if you're using it from somewhere inside the US it'll choose to display Fahrenheit instead of Celsius). You can still set the `weather_unit` variable to either `u` (for [USCS](https://en.wikipedia.org/wiki/United_States_customary_units)) or `m` (for metric) to override this behavior.
3. **Definitely don't** use our "own" `urlencode` function in bash. There are a thousand edge cases and I don't want to support all of them just for a small blocklet. Instead, **IF** (and that's a big if) the wttr.in autodetection doesn't work folks can choose to specify their location using [the `query` parameter WorldWeatherOnline supports (which wttr.in uses for their own API)](https://www.worldweatheronline.com/developer/api/docs/search-api.aspx). E.g. `São Paulo` would become `weather_location=São+Paulo` (yes, in Unicode, wttr.in supports that!), and if that doesn't return a satisfying result you can even specify a country (e.g. `London,uk`) or a UK/Canada postal/US zipcode (`90201`). Yes, this is a little more complicated for the individual user, but we're talking about highly unlikely edge-case here and I don't want to optimize a very simple script for these edge-cases.
4. Don't allow for specifying a weather label. I know this might be controversial, but wttr.in already supplies a "weatherish" label for the current conditions and I just don't see the need for adding yet another label.

I've also added fairly exhaustive tests. Obviously, they don't cover all the use-cases, but my intent is to add more of them as they appear (and break the script). For now all of the test cases should be passing.

Fixes #95 

cc @hendynz @p3k

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,3 +33,7 @@ jobs:
         shell: bash
         run: |
           tests/temp_test.sh
+      - name: Run temp test
+        shell: bash
+        run: |
+          tests/weather_test.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: |
           apt update -qq
-          apt install --no-install-recommends -y curl iproute2 bc silversearcher-ag
+          apt install --no-install-recommends -y curl ca-certificates iproute2 bc silversearcher-ag
           cp tests/xrescat_mock /usr/bin/xrescat
           cp tests/i3-msg_mock /usr/bin/i3-msg
       - name: Run net-traffic test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
         shell: bash
         run: |
           tests/temp_test.sh
-      - name: Run temp test
+      - name: Run weather test
         shell: bash
         run: |
           tests/weather_test.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: |
           apt update -qq
-          apt install --no-install-recommends -y iproute2 bc silversearcher-ag
+          apt install --no-install-recommends -y curl iproute2 bc silversearcher-ag
           cp tests/xrescat_mock /usr/bin/xrescat
           cp tests/i3-msg_mock /usr/bin/i3-msg
       - name: Run net-traffic test
--- a/scripts/weather
+++ b/scripts/weather
@@ -1,51 +1,29 @@
 #!/bin/bash
 
-# Taken from https://gist.github.com/cdown/1163649
-urlencode() {
-    # urlencode <string>
-    old_lc_collate=$LC_COLLATE
-    LC_COLLATE=C
+set -Eeux -o pipefail
 
-    local length="${#1}"
-    for (( i = 0; i < length; i++ )); do
-        local c="${1:i:1}"
-        case $c in
-            [a-zA-Z0-9.~_-]) printf "$c" ;;
-            *) printf '%%%02X' "'$c" ;;
-        esac
-    done
-
-    LC_COLLATE=$old_lc_collate
-}
-
-LABEL_ICON=${icon:-$(xrescat i3xrocks.label.weather)}
-LABEL_COLOR=${label_color:-$(xrescat i3xrocks.label.color "#7B8394")}
 VALUE_COLOR=${color:-$(xrescat i3xrocks.value.color "#D8DEE9")}
 VALUE_FONT=${font:-$(xrescat i3xrocks.value.font "Source Code Pro Medium 13")}
+BUTTON="${button:-}"
 
 # Information on the various formats: https://github.com/chubin/wttr.in
 VALUE_WEATHER_FORMAT=${weather_format:-$(xrescat i3xrocks.weather.format "1")}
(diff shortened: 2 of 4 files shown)


kgilmer (commented):
Looks like a great refactor, lots of cleanup.  I checked out your branch locally and run but see:
```bash
$ ./weather 
<span font_desc="JetBrains Mono Medium 13" color="#E6E1CF">Unknown location; please try ~33.5965685,-117.659405</span>
```

I echo the script execution if it helps in debugging:
```
$ ./weather 
++ xrescat i3xrocks.value.color '#D8DEE9'
+ VALUE_COLOR='#E6E1CF'
++ xrescat i3xrocks.value.font 'Source Code Pro Medium 13'
+ VALUE_FONT='JetBrains Mono Medium 13'
+ BUTTON=
++ xrescat i3xrocks.weather.format 1
+ VALUE_WEATHER_FORMAT=1
+ VALUE_WEATHER_FORMAT='?format=1'
++ xrescat i3xrocks.weather.units ''
+ WEATHER_UNIT=
+ '[' -n '' ']'
++ xrescat i3xrocks.weather.location ''
+ VALUE_WEATHER_LOCATION=
++ xrescat i3xrocks.weather.error_message 
+ VALUE_WEATHER_ERROR_MESSAGE=
++ curl -sS 'https://wttr.in/?format=1'
+ WEATHER='Unknown location; please try ~33.5965685,-117.659405'
+ echo '<span font_desc="JetBrains Mono Medium 13" color="#E6E1CF">Unknown location; please try ~33.5965685,-117.659405</span>'
<span font_desc="JetBrains Mono Medium 13" color="#E6E1CF">Unknown location; please try ~33.5965685,-117.659405</span>
+ '[' '!' x == x ']'
```

And, yeah when I just run the curl I get:

```bash
$ curl -sS 'https://wttr.in/?format=1'
Unknown location; please try ~33.5965685,-117.659405
```

moritzheiber:
Okay, one thing, could you visit https://wttr.in in a regular browser and post a screenshot?

And what happens if you set the `weather_location` inside the blocklet configuration?

kgilmer:
image

```bash
kgilmer@clavin:/tmp/regolith-i3xrocks-config/scripts$ weather_location="Lake Forest, CA" ./weather 
<span font_desc="JetBrains Mono Medium 13" color="#E6E1CF">☀️ +57°F</span>
kgilmer@clavin:/tmp/regolith-i3xrocks-config/scripts$ weather_location="Toronto, CA" ./weather 
<span font_desc="JetBrains Mono Medium 13" color="#E6E1CF">☁️ +37°F</span>
kgilmer@clavin:/tmp/regolith-i3xrocks-config/scripts$ ./weather 
<span font_desc="JetBrains Mono Medium 13" color="#E6E1CF">☀️ +50°F</span>
kgilmer@clavin:/tmp/regolith-i3xrocks-config/scripts$ 
kgilmer@clavin:/tmp/regolith-i3xrocks-config/scripts$ ./weather 
<span font_desc="JetBrains Mono Medium 13" color="#E6E1CF">☀️ +50°F</span>
```

It seems to be working today..?  I notice though that the default location temp seems to be different than if I specify my town specifically.  Given that we can't really depend on the service to return valid data, we need a better way of handing error cases.  I think this was tried in the past but I guess the heuristic didn't work out.

As I recall, I think there is something volatile about wttr, like a limited amount of location resolution per day based on free API usage, although I may be dreaming that up.

moritzheiber:
I've read up on wttr.in and how reliable it is and since the release of their latest API (`v3`) they seem to have resolved to rate limiting issues at least (they're not querying several services by default). Nonetheless, I've added a way of handling the "Unknown location" errors more gracefully. I don't think we can resolve them better with the current implementation.

I thought about switching away from wttr.in, to e.g. GWeather (which comes with its own slew of problems, mainly its lacking location database) or yr.no (which then needs accurate location information retrieved in the script, again), but found none of them attainable.

@kgilmer Does the script in its current state work for you? I'm going to add some documentation to the website to address the most common issues (e.g. setting your location manually)

kgilmer:
Indeed it does!  I am good w/ this change.

```bash
$ scripts/weather 
<span font_desc="JetBrains Mono Medium 13" color="#E6E1CF">☀️ +52°F</span>
```

kgilmer (approved):

PR #98, opened 2020-12-19
Fix disk-capacity button clicks, add tests

moritzheiber:
`disk-capacity` is using `sh`, which doesn't support the syntax added for the button click guard. Fixed by using `bash` instead.

Also slightly changed the way we're getting the free capacity to be less cumbersome to parse and only use `awk` instead of `grep` _and_ `awk`.

Also added tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,3 +33,7 @@ jobs:
         shell: bash
         run: |
           tests/temp_test.sh
+      - name: Run disk-capacity test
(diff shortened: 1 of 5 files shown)


kgilmer:
Looks good, thanks for the cleanup work and tests!

kgilmer (approved):

eblume:
I just saw this in your reply to my slack post, @moritzheiber  - thank you so much!

PR #99, opened 2020-12-21
Use awk instead of bash printf to avoid octal conversion

moritzheiber:
bash has this weird way of converting numbers with leading zeros to octals, which means that the `todo` blocket will throw an error whenever _just_ a zero is passed into it. This works around it by using `awk` to `printf` the statement.

Also added tests to verify the script gives us proper output.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,3 +37,7 @@ jobs:
         shell: bash
         run: |
           tests/disk-capacity_test.sh
+      - name: Run todo test
+        shell: bash
+        run: |
+          tests/todo_test.sh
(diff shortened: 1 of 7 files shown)


cheginit:
So that's what it is. It seems that `td count` returns the value with some additional characters so here's how I worked around it:
```bash
ntask=$(td count $FILTER | sed 's/[^0-9]*//g')
COUNT="$(printf ' %-2s' "${ntask: :-1}")"
```
But your solution is more elegant.
Another [PR](https://github.com/regolith-linux/regolith-todo/pull/4) is being merged, can you please wait for a bit till we merge that one then rebase?
BWT, thanks for the test, it was needed!

moritzheiber:
Sure thing, no problem.

cheginit:
We merged that PR, so please rebase.

PR #101, opened 2020-12-28
Add padding for net-traffic, extensive mocked tests

moritzheiber:
As requested, I've added a padding of 4 characters (plus the unit and an extra space in the `<span>` output) so the output on the bar isn't jittering whenever units and measurements are changing.

I've also completely rewritten the tests to actually use mocked values (using `umockdev`) instead of trying to come up with a magical regular expression.

The script should break much less now, especially on upgrades.


moritzheiber:
I believe I forgot to push two extra commits that are showing up in this PR now. They are safe to merge as well though (adding a line feed and `.editorconfig`).

conf.d/30_net-traffic
```diff
@@ -1,3 +1,3 @@
 # Displays network load
 [net-traffic]
-interval=repeat
\ No newline at end of file
+interval=repeat
```
kgilmer, line 3: Ah, thanks for rolling this in.

tests/net-traffic_test.sh
```diff
@@ -5,16 +5,155 @@ set -Eeux -o pipefail
 SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
 cd "${SCRIPT_DIR}/"
 
-# Regular syntax
-[ "2" -eq "$(../scripts/net-traffic | ag -o '(?<=\>\s)(\d{1,4}|\d{1,3},?\d{0,1}?)[GMKB](?=\<)' | wc -l)" ]
+SYS_PREFIX="/sys/class/net"
```
kgilmer, line 8: Wow this is quite a test!

kgilmer (approved):
Beautiful, thanks for setting the padding and adding all of the test code as well!

PR #110, opened 2021-01-05
Filter VPN interfaces from net-traffic scripts

### LINKED ISSUE #109: net-traffic fails when connected to VPN and causes CPU load
While trying to figure out why I get a continuous 15% CPU load without any applications running, I found that it was because the `net-traffic` script continuously fails.

Running it by hand gives:

    » /usr/share/i3xrocks/net-traffic
    cat: /sys/class/net/static/statistics/rx_bytes: No such file or directory

Indeed there is no `static` entry in `/sys/class/net`:

    » ls /sys/class/net
    br-adad92378979  docker0  lo  ppp0  virbr0  virbr0-nic  wlp59s0

The mistake happens in the line `IF="${BLOCK_INSTANCE:-$(ip route show default | awk 'NR==1 { print $5 }')}"`, where `ip route` gives:

    » ip route show default
    default dev ppp0 proto static scope link metric 50 
    default via 192.168.0.1 dev wlp59s0 proto dhcp metric 600

So it picks up the wrong column, or the wrong line. Changing the line in the script to the following makes it work and puts my CPU at ease both with and without being connected to the VPN:

    IF="${BLOCK_INSTANCE:-$(ip route show default | grep 'via' | awk 'NR==1 { print $5 }')}"

However I don't know if that is a good solution in the general case.

moritzheiber:
By default, VPN interfaces are showing up as default routes on the `main` table because they can be set as the default destination for traffic, while the actual, physical device (e.g. `eth0`) sits back as a sort of "shadow" default device (which is still necessary, otherwise you wouldn't be able to talk to your router/the Internet anymore). Unfortunately, this can cause the command `ip route show default` (which queries the  `main` table) to return multiple entries, including the VPN-bound default routes. Previously we tried to work around this by just cutting the output to a single line and taking whatever was returned. Unfortunately, that's not reliable enough (see the `ip` mocks I added for both, `tun0` and `ppp0` default routes) as the output slightly differs and makes parsing the results cumbersome.

I've decided to fix this by hard-filtering the output for any known VPN prefixes (`(ppp|tun|tap)`) which should then return only a single line, which is also the "true" default route. Traffic measurements should still work correctly, as any traffic passing through VPN tunnels is also aggregated into the device statistics of the relevant underlying network interface.

Fixes #109 

_PS: Yes, there can be multiple distinct default routes, but not in the same table and usually only on servers. We'll address that edge case once it hits us :wink:_

### DIFF
--- a/scripts/net-traffic
+++ b/scripts/net-traffic
@@ -15,7 +15,7 @@ RT=${RT:-}
 NUMFMT="numfmt --to iec --format %f --padding 5"
 
 # determine the net interface name
-IF="${BLOCK_INSTANCE:-$(ip route show default | awk 'NR==1 { print $5 }')}"
+IF="${BLOCK_INSTANCE:-$(ip route show default | awk '!/(ppp|tun|tap)/ { print $5 }')}"
 [ "$IF" = "" ] && exit
 
 IF_PATH="/sys/class/net/${IF}"
--- a/tests/fixtures/net-traffic/first/ip
+++ b/tests/fixtures/net-traffic/first/ip
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+echo "default via 10.0.0.1 dev tun0 proto static metric 50
+default via 192.168.1.1 dev eth0 proto dhcp metric 600"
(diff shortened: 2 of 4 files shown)


kgilmer (approved):
LGTM

PR #111, opened 2021-01-10
net-traffic cpu load w/ no or invalid network devices fix

kgilmer:
Capture error cases involving invalid network device paths and sleep before returning to resolve CPU consuming loop due to  interval type of i3blocks.  Addresses https://github.com/regolith-linux/regolith-desktop/issues/532.

## Testing Done

1. On local machine, observe issue when disabling wifi.  
2. Replace `release` version of `net-traffic` script with this
3. disable and renable wifi
4. observe no obvious increase load on machine based on network connectivity state

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,14 @@
+regolith-i3xrocks-config (3.5.5-1ubuntu1) bionic; urgency=medium
+
+  [ Moritz Heiber ]
+  * Added gawk specific testing for temp script (#108)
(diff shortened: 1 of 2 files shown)


ploum:
I confirm that the fix solves the issue for me.

ploum (approved):

moritzheiber (approved):
I would like to come up with a test for this ideally but I don't have the mindspace for it right now. I guess it'll work as is.

Issue #105, opened 2020-02-16
Add microphone block


gnzlbg:
Add a microphone block, that shows up if the microphone is muted or not, and that lets you control the microphone volume.

nicolas-lopez:
Hello I would like to submit this script for microphone handling: https://pastebin.com/xSYVzY93
Unfortunately I don't have time to make a proper pull request

kgilmer:
Thanks @nicolas-lopez !  I've created a PR from your paste.  https://github.com/regolith-linux/regolith-i3xrocks-config/pull/117

jc00ke:
This can be closed, yeah?

kgilmer:
Yep, thanks @jc00ke

PR #117, opened 2021-02-17
microphone status indicator from Niclas Lopez

kgilmer:
Take nicolas-lopez's contribution via gist as microphone status indicator.  Related to https://github.com/regolith-linux/regolith-i3xrocks-config/issues/105.

### DIFF
--- a/scripts/microphone
+++ b/scripts/microphone
@@ -1,0 +1,101 @@
+#!/bin/bash
+
+# Copyright (c) 2020 Guillaume Deflaux, Marco Buzzanca
+#
+# Based on the work of:
+#   2014 Julien Bonjean <julien@bonjean.info>
+#   2014 Alexander Keller <github@nycroth.com>
+#
+# GNU GENERAL PUBLIC LICENSE
+#    Version 3, 29 June 2007
+#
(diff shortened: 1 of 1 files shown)


cheginit (changes requested):
Just a minor thing, use JetBrains Mono as the default font instead of Source Code Pro.

moritzheiber:
Sorry for not responding earlier .. as far as I can see this is a carbon copy of the volume indicator .. with just a few adjustments for displaying a microphone emoji instead of the volume?

As far as I remember we discussed reworking the volume blocklet the last time and came to the conclusion that it should rather be ported to Pulseaudio. I may be remembering this incorrectly though.

jc00ke:
Is this PR even necessary anymore? Looks like this script [already exists](https://github.com/regolith-linux/regolith-i3xrocks-config/blob/f5fcb40ff123b3d8d45e579f6146acee7d988854/scripts/microphone) minus the default font, which most are `Source Code Pro` anyway (only `memory` and `media-player` default to `JetBrains Mono`).

kgilmer:
Yes this should be closed, thanks @jc00ke !

PR #126, opened 2022-05-11
Use /proc/net/route for interface detection

### LINKED ISSUE #125: net-traffic exits unexpectedly
Hello,
on 20.04 with a default route through a VPN the IF detection in https://github.com/regolith-linux/regolith-i3xrocks-config/blob/master/scripts/net-traffic#L19 seems to fail:

```
$ ip route show default                                      
default dev vpn0 proto static scope link metric 50 
default via 192.168.178.1 dev enxb07b2540cf7b proto dhcp metric 10

$ ip route show default | awk '!/(ppp|tun|tap)/ { print $5 }'
static
enxb07b2540cf7b
```

I've not deeply looked into this, but I think this should work:

```
--- /tmp/net-traffic    2022-01-17 09:16:22.685425820 +0100
+++ /usr/share/i3xrocks/net-traffic     2022-01-17 09:16:08.865327514 +0100
@@ -16,7 +16,7 @@
 NUMFMT="numfmt --to iec --format %f --padding 5"
 
 # determine the net interface name
-IF="${BLOCK_INSTANCE:-$(ip route show default | awk '!/(ppp|tun|tap)/ { print $5 }')}"
+IF="${BLOCK_INSTANCE:-$(ip route show default | awk '!/(ppp|tun|tap|vpn)/ { print $5 }')}"
```

Now the problem is that I think this is not a good solution. (Thus this is an issue and not a PR)

For example with wireguard the interface name could be completely custom, cf. https://www.wireguard.com/netns/ - although I've seen people use the `wg0` from the examples, but `wg-quick` again derives it from the config.

In any case, I think there should be a `| head -n 1` at the end, because a 0 as an output value might be better than a spammy log file like I have it:

```
Jan 17 08:39:05 ws6 regolith.desktop[2724]: <3>[net-traffic] Command '/usr/share/i3xrocks/$BLOCK_NAME' exited unexpecte>
Jan 17 08:39:15 ws6 regolith.desktop[2709]: <3>[net-traffic] Command '/usr/share/i3xrocks/$BLOCK_NAME' exited unexpecte>
Jan 17 08:39:15 ws6 regolith.desktop[2695]: <3>[net-traffic] Command '/usr/share/i3xrocks/$BLOCK_NAME' exited unexpecte>
...
```

winks:
Fixes #125

### DIFF
--- a/scripts/net-traffic
+++ b/scripts/net-traffic
@@ -15,8 +15,9 @@ RT=${RT:-}
 # Padding the output to 5 characters to avoid jittering on the bar
 NUMFMT="numfmt --to iec --format %f --padding 5"
 
-# determine the net interface name
-IF="${BLOCK_INSTANCE:-$(ip route show default | awk '!/(ppp|tun|tap)/ { print $5 }')}"
+# determine the net interface name, but only use one interface
+# there could be several default routes, if you have a VPN for example
+IF="${BLOCK_INSTANCE:-$(awk '$2 == 00000000 && $1 ~ /^(wl|en|eth)/ { print $1; exit }' < /proc/net/route)}"
 
 IF_PATH="/sys/class/net/${IF}"
 


kgilmer:
Building in https://github.com/regolith-linux/voulage/actions/runs/2339243714

PR #127, opened 2022-05-29
fixed xml escaping

ferdinandyb:
Adding xml escaping (plus some formatting that was auto-done by isort and black which I left in).

### DIFF
--- a/scripts/focused-window-name
+++ b/scripts/focused-window-name
@@ -22,12 +22,20 @@
 # from command line arguments, but the latter takes precendence.
 #
 #
-import sys
 import argparse
 import os
 import subprocess
+import sys
 
 from i3ipc import Connection, Event
+
+
+def escape(name):
+    name = name.replace("&", "&amp;")
+    name = name.replace("<", "&lt;")
+    name = name.replace(">", "&gt;")
+    name = name.replace('"', "&quot;")
+    return name
 
 
 def value_span(value):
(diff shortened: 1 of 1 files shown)


scripts/focused-window-name
```diff
@@ -22,37 +22,58 @@
...
+).communicate()
 value_color = value_color.decode("utf-8")
 label_color, err = subprocess.Popen(
-    ['xrescat', 'i3xrocks.label.color', '#7B8394'], stdout=subprocess.PIPE).communicate()
+    ["xrescat", "i3xrocks.label.color", "#7B8394"], stdout=subprocess.PIPE
+).communicate()
 label_color = label_color.decode("utf-8")
 value_font, err = subprocess.Popen(
-    ['xrescat', 'i3xrocks.value.font', 'Source Code Pro Medium 13'], stdout=subprocess.PIPE).communicate()
+    ["xrescat", "i3xrocks.value.font", "Source Code Pro Medium 13"],
```
kgilmer, line 74: (minor) can you update the default font to simply be `sans`?

kgilmer (approved):
LGTM

scripts/focused-window-name
ferdinandyb: Sure. Without any font size specs, just `sans`?

kgilmer (approved):

cwh1te:
I was going to open an issue until I found this patch. The bug has bugged me for awhile. Thanks for patching it!

In troubleshooting, I found out that GLib already has an escaping function, so it may be wise to use `gi.repository.GLib.markup_escape_text()` rather than a custom function. That way if the template language ever adds or removes escaped characters, this code should automatically be compatible (I'm scratching my head wondering how single quotes don't break this currently).

Any idea whether this bug could have security implications? I'm not familiar enough with Pango to say one way or the other, but it seems possible.

PR #129, opened 2022-06-17
Adding a fix to not show anything on empty workspace

ferdinandyb:


kgilmer (approved):
Wow, nice cleanup as well.  Looks good!

PR #130, opened 2022-06-20
Remove extra space in nm-vpn output

jjsearle:
There are currently two space characters between the icon and the name of the VPN connection.

All other scripts only have a single space. This removes the extra space to make it consistent with the other scripts.


kgilmer (approved):
LGTM

PR #131, opened 2022-06-24
Add updates block

jjsearle:
This adds a block to check for package updates

### Updates available

updates-available

In this screenshot there are a total of 12 updates available, 3 of which are security updates

### Reboot required
reboot-required

### Xresources

- `i3xrocks.updates.checkcmd` : command to run to get the number of updates should return in format '<total_updates>;<security_updates>'
- `i3xrocks.updates.displayzero` : bool, if true display even if there are zero updates
- `i3xrocks.action.updates` : command to run when block is left clicked
- `i3xrocks.action.updates.reboot` : command to run when block is left clicked and reboot is required
- `i3xrocks.label.updates` : glyph icon to display
- `i3xrocks.label.updates.reboot` : glyph icon to display when reboot is required
- `i3xrocks.updates.interval` : interval in seconds between checks

### DIFF
--- a/conf.d/80_updates
+++ b/conf.d/80_updates
@@ -1,0 +1,3 @@
+# Package Updates
+[updates]
+interval=xresource:i3xrocks.updates.interval 3600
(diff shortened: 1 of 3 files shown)


kgilmer (approved):
LGTM

Issue #133, opened 2022-07-01
Update indicator over eager to reboot system


kgilmer:
The new update indicator shows a user what packages need to be updated. It has a mode that informs if some updated package requires a reboot to take effect, such as a Kernel update.  However when clicking on that in the bar the system is immediately rebooted.  It seems from experience that people sometimes just randomly click things and so we would not want the system to just suddenly reboot.  The ideal experience IMO would be to copy the pattern of the reboot hotkey in i3:

```
/usr/share/regolith/i3/config.d/55_session_keybindings:set_from_resource $i3-wm.program.reboot i3-wm.program.reboot /usr/bin/gnome-session-quit --reboot
```

jjsearle:
Hey Ken, I agree, much better behaviour. See the PR above that fixes this.

PR #134, opened 2022-07-01
replace reboot command with saner default

jjsearle:
Addresses https://github.com/regolith-linux/regolith-i3xrocks-config/issues/133


kgilmer (approved):
LGTM, thanks @jjsearle !

PR #137, opened 2022-07-25
Fixed volume and microphone mixer_value to display values below 10 and prepend a zero if value is below 10

### LINKED ISSUE #135: i3xrocks-volume - volume below 10% not shown
Hey,

I'm currently testing regolith with an virtual minimal Ubuntu installation (like it's described here https://regolith-desktop.com/docs/howtos/minimal-ubuntu-install/) and figured out that if I change the volume below 10% the volume is not displayed anymore:

image

If the volume is >= 10% everything is fine:

image

The same behavior applies for the microphone volume.

The reason seems to be the `sed` regex which is used within the `mixer_value` function: https://github.com/regolith-linux/regolith-i3xrocks-config/blob/7bb3464cc56ec3e42830e9126a246769a86d3bba/scripts/volume#L66

I think it would be better to only filter on a single number and add a `0` prefix in case the value is less than 10, for example:

```bash
mixer_value() {
    VAL=$(mixer_info | sed -n -E "s/.*\[(-?[0-9]+)$UNIT\].*/\1/p" | head -n 1)
    if [ -z "$VAL" ]; then
        # return empty string of same width to prevent bar shift.
        echo "   "
    elif [ "$VAL" -lt "10" ]; then
        echo "0$VAL$UNIT"
    else
        echo "$VAL$UNIT"
    fi
}
```
So volume below 10% is also displayed as expected:

image

sdrwtf:
## Description
This PR fixes #135, which describes the issue that if the volume or microphone level is below 10 no value is displayed in the status bar.

## Changes
Altered the `sed`-regex within the `mixer_value` function to also include single-digit values and extended the if-logic to prepend a zero if the value s below 10 to avoid bar shift / jumping output in the volume and microphone script.

## Test
Successfully tested both scripts in a virtual machine (Ubuntu server 22.04 LTS with regolith set up like it's described here: [https://regolith-desktop.com/docs/howtos/minimal-ubuntu-install/](https://regolith-desktop.com/docs/howtos/minimal-ubuntu-install/)).

### DIFF
--- a/scripts/microphone
+++ b/scripts/microphone
@@ -63,10 +63,13 @@ mixer_status() {
 }
 
 mixer_value() {
-    VAL=$(mixer_info | sed -n -E "s/.*\[(-?[0-9]+.?[0-9]+)$UNIT\].*/\1/p" | head -n 1)
+    VAL=$(mixer_info | sed -n -E "s/.*\[(-?[0-9]+)$UNIT\].*/\1/p" | head -n 1)
     if [ -z "$VAL" ]; then
         # return empty string of same width to prevent bar shift.
         echo "   "
+    elif [ "$VAL" -lt "10" ]; then
+        # prepend a zero if volume is below 10
+        echo "0$VAL$UNIT"
     else
         echo "$VAL$UNIT"
     fi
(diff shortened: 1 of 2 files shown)


kgilmer:
LGTM, thanks @sdrwtf !

PR #143, opened 2022-10-07
Use a … instead of ... to show more of the artist and title

jeffothy:
When we truncate the media-player's artist and track title it would be nice to show as much of it as possible.

Previously we chop 4 and 5 characters off of the artist and track respectively for the spacing and ellipsis as `...`.
This PR simply uses the unicode ellipsis character `…` U+2026 so we only have to chop 2 and 3 characters off!


kgilmer:
Nice, sounds like a good improvement @jeffothy .  Before I merge this, can you mention briefly what did to test your work?

kgilmer (approved):
LGTM

PR #145, opened 2022-12-19
fix(keyboard-layout): open keyboard menu instead

Tooa:
Hi :wave:,

I installed Regolith Linux 2.2 from the ISO [here](https://regolith-desktop.com/#get-regolith-22) today. After inspecting the `i3xrocks-keyboard-layout` package, I found that the left click action links to the wrong gnome settings menu:

before

From my understanding it makes more sense to link to the keyboard layout settings here:

after

I assume the menu setting entries were changed upstream. Correct me if I am wrong. Let me know if you want anything changed.


kgilmer:
Fantastic, thanks @Tooa !

PR #148, opened 2023-07-31
Use value colors for battery percentage and charge

pietermarsman:
I think this fixes a layout error. All i3xrocks indicator values are shown with `xrescat i3xrocks.value.color` except for battery. The battery script uses `xrescat i3xrocks.label.color`. This change fixes it. 

I've made this change for my own setup and think maybe others will like it too.


kgilmer (approved):
LGTM

kgilmer:
Looks good, thanks @pietermarsman .  This will ship w/ Regolith 3.0.

PR #157, opened 2025-05-18
volume: avoid logging error while detecting pulseaudio

### LINKED ISSUE #154: i3xrocks-volume: "pactl: command not found"
`journalctl --follow` shows a slow steady stream of:

```
regolith-session-x11[4037999]: /usr/share/i3xrocks/scripts/volume: line 38: pactl: command not found
```

I _upgraded_ to 24.04 rather than doing a full install. As such, it's possible my system is in a defunct state.

That being said, the [volume script](https://github.com/regolith-linux/regolith-i3xrocks-config/blob/master/scripts/volume#L38) appears to assume `pactl` is installed.

Unlike #152, my volume icon is working and can change my audio levels. Uninstalling `i3xrocks-volume` eliminates the noise in logs.

mpareja:
The volume script attempts to detect pulseaudio by calling `pactl`. On machines without `pactl` installed, the detection generates a steady stream of noise in the logs, `journalctl --follow` shows:

```
regolith-session-x11[4037999]: /usr/share/i3xrocks/scripts/volume: line 38: pactl: command not found
```

Closes #154.

### DIFF
--- a/scripts/volume
+++ b/scripts/volume
@@ -35,7 +35,7 @@ UNIT="${2:-$(xrescat i3xrocks.volume.unit %)}"
 # For ALSA users, you may use "default" for your primary card
 # or you may use hw:# where # is the number of the card desired
 MIXER="default"
-pactl list 1>/dev/null && MIXER="pulse"
+pactl list 1>/dev/null 2>&1 && MIXER="pulse"
 lsmod | grep -q jack && MIXER="jackplug"
 MIXER="${3:-$(xrescat i3xrocks.volume.mixer $MIXER)}"
 


kgilmer:
Hi @mpareja , thanks for the fix and clear description.  Looking at the script, I wonder if `pactl` is an optional or required dependency.  The script looks to me like it's required, and if so, I suggest this fix instead, which simply requires that the package providing `pactl` is installed as well ~

```
diff --git a/debian/control b/debian/control
index fb287fd..04b4252 100644
--- a/debian/control
+++ b/debian/control
@@ -140,7 +140,8 @@ Package: i3xrocks-volume
 Architecture: any
 Depends: ${misc:Depends},
     regolith-i3xrocks-config,
-    alsa-utils
+    alsa-utils,
+    pulseaudio-utils
 Description: Indicator to show volume.
  An i3xrocks indicator for showing the current sound volume.
```

LMK what you think :)

mpareja:
From what I can tell, `pactl` is _only_ used to detect which device to manipulate using `amixer` (which is a required dependency). I don't think we should require installing the pulse audio tools.

kgilmer (approved):
LGTM

PR #160, opened 2026-04-04
fix: align ilia launch commands with wm config

anatolepain:
This PR makes the ilia launch commands match the i3/sway wm configs ([i3](https://github.com/regolith-linux/regolith-wm-config/blob/main/usr/share/regolith/i3/config.d/20_ilia#L4), [sway](https://github.com/regolith-linux/regolith-wm-config/blob/main/usr/share/regolith/sway/config.d/20_ilia)).

- Use scripts that read `ilia.stylesheet` from Xresources and pass `-t` when the stylesheet file exists, otherwise let ilia fall back to its default style (from `default_css` in [Application.vala](https://github.com/regolith-linux/ilia/blob/main/src/Application.vala#L4)).
- Run `ilia -p apps` for the app launcher and `ilia -a -p keybindings` for the keybindings viewer, consistent with the wm configs.

### DIFF
--- a/conf.d/20_app-launcher
+++ b/conf.d/20_app-launcher
@@ -2,4 +2,3 @@
 #
 [app-launcher]
 full_text=xresource:i3xrocks.label.app_launcher
-command=i3-msg -q "exec --no-startup-id ilia -a -p apps"
--- a/conf.d/20_info
+++ b/conf.d/20_info
@@ -1,4 +1,3 @@
 # Link to execute shortcuts.
 [info]
 full_text=xresource:i3xrocks.label.help
-command=i3-msg -q "exec --no-startup-id ilia -p keybindings"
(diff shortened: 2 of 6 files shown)


kgilmer:
Hi @anatolepain , thanks this looks like a good improvement.  Can you briefly mention how you tested?

anatolepain:
Hi, I tested this locally on Regolith 3.4 with Ubuntu Noble 24.04 LTS.

I manually applied the changes by editing the configs in `/usr/share/i3xrocks/config.d/` and adding the scripts in `/usr/share/i3xrocks/scripts/`. I then reloaded the config with `mod + shift + r` or logged out and back in, and tested the app launcher and info icons from the bar.

I checked the following cases:

- `ilia.stylesheet` set in Xresources and pointing to an existing custom stylesheet
- `ilia.stylesheet` set but pointing to a missing file
- the default Regolith `ilia.stylesheet` value
- `ilia.stylesheet` removed from Xresources
- `ilia.stylesheet` removed and the default `ilia.css` temporarily moved away

I verified the Xresources value with `xrdb -query | grep ilia.stylesheet`.

